PR #1558, opened 2014-11-05
Use bundler for installing gems for CSS compiling.

dracos:
This might assume that rubygems is installed, but it did that before anyway.


mhl:
I've rebased this onto master and added ruby-bundler to the packages installed by site-specific-install. (We're no longer supporting Precise.) So I think this is good to merge :+1: Sorry about the delay!

Issue #1492, opened 2014-07-14
ZA: questions page improvements


mhl:
PMG have requested several improvements to the questions page - we should break these out into new issues:

- [ ] Show a list of "most recently answered questions" at the top of `/questions`
- [ ] Pagination for ministries with many questions
- [ ] A snippet of the question text to be included on the question list pages, more like: http://myreps.sailing360.co.za/questions
- [ ] Remove the next and previous ministries links (they're a lot of noise if there are only a couple of questions.
- [ ] There are some duplicate ministries on the questions list page - they should be merged
- [ ] Perhaps make it more obvious how to get back to the index of questions (at the moment the link in the breadcrumbs isn't very obvious)

paullenz:
closing as consolidating points in - https://github.com/mysociety/pombola/issues/1797

Issue #1561, opened 2014-11-12
ZA: formatting of answers, particularly those including tables, needs improvement


geoffkilpin:
An answer such as that at http://www.pa.org.za/questions/questions-asked-to-the-minister-in-the-presidenc-5/nw2022e-19-september-2014#s644044 is difficult to read due to browsers ignoring its plain text formatting (which in turn is an output of antiword). This is especially the case where an answer includes a table.

This could potentially be addressed by using the Django [linebreaks filter](https://docs.djangoproject.com/en/1.7/ref/templates/builtins/#std:templatefilter-linebreaks), although this would result in line breaks within paragraphs.

The displayed text is:

```[pic] MINISTER IN THE PRESIDENCY: REPUBLIC OF SOUTH AFRICA Private Bag X1000, Pretoria, 0001, Union Buildings, Government Avenue, PRETORIA Tel: (012) 300 5200, Website: www.thepresidency.gov.za NATIONAL ASSEMBLY QUESTION FOR WRITTEN REPLY QUESTION NUMBER: 1661 Mr S J Masango (DA) to ask the Minister in The Presidency: Whether any Ratel Infantry Combat Vehicles have been exported in each specified calendar year between 2003 and 2014; if not, in each specified calendar year, why not; if so, in each specified calendar year, what was the (a) name of the importing state, (b) name of the end user, (c) quantity of vehicles exported and (d) total value of the vehicles exported? REPLY: 1. The details of Ratel Infantry Combat Vehicles exported over the period 2003 to 2014 under the NCACC authority are as reflected in the table below. |YEAR |COUNTRY & END USER |QUANTITY | | | | | |2003 |Ghana Government |19 | | |Jordan Government |100 | | | | | |2004 |Ghana Government |20 | | |Jordan Government |221 | | | | | |2007 |Rwanda Government |35 | | | | | |2008 |Senegal Government |20 | | | | | |2012 |Zambia Government |29 | | | | | Approved / Not approved Mr J T Radebe, MP Minister in the Presidency Date:```

While the plain text is:

```[pic]

            MINISTER IN THE PRESIDENCY: REPUBLIC OF SOUTH AFRICA
   Private Bag X1000, Pretoria, 0001, Union Buildings, Government Avenue,
                                  PRETORIA
           Tel: (012) 300 5200, Website: www.thepresidency.gov.za


NATIONAL ASSEMBLY

QUESTION FOR WRITTEN REPLY

QUESTION NUMBER: 1661


[...]

PR #1562, opened 2014-11-12
ZA: New question and answer landing page

geoffkilpin:
Implements the first stages of #1492. This is largely intended to enable work on the design of the new landing page and further work to ```views.py``` is probably necessary.

### DIFF
--- a/pombola/south_africa/static/sass/_south-africa_overrides.scss
+++ b/pombola/south_africa/static/sass/_south-africa_overrides.scss
@@ -2572,3 +2572,19 @@ ul.pager {
         }
     }
 }
+
+.questions {
+    .intro {
+        font-weight:bold;
+    }
+    .question {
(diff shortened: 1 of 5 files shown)


geoffkilpin:
I've updated this to include @zarino's great design work from #1564.

One issue that may need discussion is how I've addressed #1561. In order to maintain table formatting, it is necessary to use a monospace font and reduce the font size to 14px (to avoid text wrapping on an ordinary computer screen). This doesn't look particularly attractive, but seems better than the confusing text we currently see with tables.

mhl:
@geoffkilpin Sorry that this has taken so long to deal with. I think it's best to just deploy this as is, since it's a great improvement anyway, and then deal with remaining issues. (e.g. the monospacing and whitespace preservation isn't great on mobile) Many thanks for all your work on this. :+1:

PR #1590, opened 2015-01-15
ZA Internet explorer fixes

chrismytton:
This updates django-pipeline to a version which uses `text/javascript` in the type attribute on script tags. It fixes the carousel in IE 8.

Fixes #1577

Issue #1577, opened 2014-12-02
Homepage carousel on pa.org.za not working on IE 11 & 8


paullenz:
![unnamed](https://cloud.githubusercontent.com/assets/909811/5261658/f9bd4b84-7a13-11e4-87a0-b60254e298e5.jpg)
![unnamed1](https://cloud.githubusercontent.com/assets/909811/5261660/fad6a5ce-7a13-11e4-8b84-56df96aa625f.jpg)

PR #1591, opened 2015-01-15
ZA Internet Explorer fixes part 2

chrismytton:
This follows on from #1590 and fixes some javascript errors which were only occurring in production when django-pipeline was active.

Part of #1577


mhl:
:+1:

PR #1596, opened 2015-01-19
Suppress the warnings about missing loggers for elasticsearch

mhl:
If there are no loggers defined for the elasticsearch Python package you
get errors like the following:

  No handlers could be found for logger "elasticsearch" (in production)
  No handlers could be found for logger "elasticsearch.trace" (local dev)

(I'm not sure why there's that difference in those logger names.)

This commit sets up loggers for both names in the Django settings module
that will send any logging of level INFO and above to standard error.
This suppresses that warning and should mean that we're able to see
real warnings and errors in the logs.

### DIFF
--- a/pombola/settings/base.py
+++ b/pombola/settings/base.py
@@ -244,6 +244,16 @@ LOGGING = {
             'level': 'ERROR',
             'propagate': True,
         },
+        'elasticsearch': {
+            'handlers': ['stream_to_stderr'],
+            'level': 'INFO',
+            'propagate': True,
+        },
+        'elasticsearch.trace': {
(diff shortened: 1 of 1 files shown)


davea:
It appears that the logging message being generated and causing the error is:

    List of nodes should not include schema information (http://): 'http://127.0.0.1:9200'.

This was fixed in elasticsearch/elasticsearch-py#149 and the latest version of `elasticsearch` on pypi (1.3.0, pombola uses 1.2.0 right now) doesn't exhibit the log warning behaviour.

In any case, this PR LGTM :+1:

PR #1598, opened 2015-01-21
Some changes to reduce the number of queries Pombola makes

mhl:
This series of commits make various changes to reduce the huge number
of queries that some pages on Pombola cause.


chrismytton:
Looks good to me! :+1: :speedboat:

Issue #1505, opened 2014-08-08
Adding information on Bills to the Mzalendo site


paullenz:
In part replacing: https://github.com/mysociety/pombola/issues/1440

Content from prior ticket:

I have been doing a bit more digging around the area of bills for the website - in brief:

Kenya law looks like it has a good archive of who has been associated with various bills and PDFs that could be linked out to: http://kenyalaw.org/kl/index.php?id=4248

however it doesn't show people who have contributed to the bill other than the sponsor - I am struggling to find more details on this - see later also

there is a similar (but less easily parable as a PDF? - or rather multiple PDFs) - tracker on the gov site - http://www.parliament.go.ke/index.php/the-senate/house-business/bills-tracker

the actual bills themselves on the gov site are pretty poorly scanner PDFs - http://www.parliament.go.ke/plone/senate/business/bills-1/public-appointments-county-assemblies-approval-bill-2014/view

So I guess the main question is, when it comes to linking people to bills:

1) Is it okay to just have sponsors?
2) If the answer to 1) is "no" then what are your thoughts on getting hold of the additional information?

Update from Jessica on the Bill question:

Just the name of the bill and sponsors would be enough. Parliament does not have good copies of the bills. www.kenyalaw.org which is run by the Judiciary is the authoritative site for all legal info.

davea:
I've made some progress on this, here's what it looks like right now on a profile page:

![screen shot 2015-01-20 at 14 27 56](https://cloud.githubusercontent.com/assets/4776/5818856/c5dc856e-a0b0-11e4-961a-9473c20e3ef3.png)

Each bill links directly to the PDF on kenyalaw.org. Am I on the right track?

JessMusila:
Yes, on the right track. Where the bill has already been assented to link
to the Act.

davea:
The profile page of a politician now shows only the 10 most recently sponsored bills:

![screen shot 2015-01-21 at 15 05 53](https://cloud.githubusercontent.com/assets/4776/5838553/2511b340-a180-11e4-9a7f-52a8850f8f30.png)

with a link to a page displayed all of their sponsored bills:

![screen shot 2015-01-21 at 15 06 57](https://cloud.githubusercontent.com/assets/4776/5838569/37e867ac-a180-11e4-91f9-7f70209992bb.png)

Those who have sponsored fewer than 10 bills and those who have never sponsored a bill aren't left out:

[...]

PR #1599, opened 2015-01-22
Add National Assembly and Senate bills to Mzalendo

davea:
This PR:

 * Adds bills that a politician has sponsored, if any, to the bottom of their profile page
 * Adds pages to view all bills in each parliamentary session
 * Adds a new menu item to the 'Parliament' top-level nav to link to the above lists
 * Enables bills to be edited in the admin
 * Provides a management command, `kenya_scrape_bills` to populate the database with bill info from kenyalaw.org


pombola/kenya/templates/core/person_bills.html
```diff
@@ -0,0 +1,32 @@
...
+{% block title %}{{ object.name }} Bills Sponsored{% endblock %}
+
+{% block subcontent %}
+
+  {% if settings.ENABLED_FEATURES.bills %}
+    {% with object.bills_sponsored.count as bills_count %}
+      {% if bills_count %}
+        <h2>Bills Sponsored</h2>
+
+        <p>{{ object.name }} has sponsored {{ bills_count }} bills.</p>
```
mhl: A very small thing, but this could be `bill{{ bills_count|pluralize }}` so we don't get "has sponsored 1 bills".

mhl:
Looks good! I think all the things I've commented on a pretty small.

davea:
Thanks for the feedback. I've squashed some commits and tidied up those few things you mentioned throughout the code. Are you happy for this to be merged?

Regarding the running of the management command via cronjob: it was my intention that the command would be a one-off to get the bulk of the existing bills loaded in, and subsequent bills would be added by hand via the admin. This seems to be what the comments in #1505 suggest. If the workflow would benefit from regularly running the import command then I'll modify it to suit.

mhl:
@davea Great, :+1: - feel free to merge and deploy

PR #1609, opened 2015-01-28
Don't convert empty tags to self-closing tags in raw HTML blog posts

### LINKED ISSUE #1608: When raw content blog posts are rendered, empty tags are converted to self-closing
This creates some problems, e.g. a video embedded with: `<iframe></iframe>` will be rendered as `<iframe/>` which then apparently eats any following paragraph...

mhl:
There was a problem that content after embedded videos (included via
YouTube's suggested <iframe ...></iframe> HTML) wasn't being displayed.
This turned out to be because we pass the raw HTML through lxml to
ensure that the output is well-formed, but this was converting empty
tags to self-closing tags and <iframe .../> and that eats the following
content in the page.

Adding method="html" to lxml.etree's tostring function preserves the
separate close tag.

Fixes #1608

### DIFF
--- a/pombola/info/models.py
+++ b/pombola/info/models.py
@@ -136,7 +136,7 @@ class InfoPage(ModelBase):
             # Parsing the HTML with lxml and outputting it again
             # should ensure that we have only well-formed HTML:
             parsed = lxml.html.fromstring(self.raw_content)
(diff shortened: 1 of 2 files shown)


davea:
Looks good to me! :+1:

PR #1618, opened 2015-02-13
ZA: Home page blog fixes

### LINKED ISSUE #1571: Blog Display Changes to pa.org.za
Three specific changes: 

1) Is it possible to order the blog posts horizontally (2 per row) instead of vertically on the homepage. For example, if we upload two blogs today, they should appear next to each other as opposed to above and below: (see sketch below)
 
2 When we click on See All link (on the homepage just under That Week in Parliament), we are directed to a page that has the blog posts from all the categories. We only want to view the That Week in Parliament (and merged Impressions) blogs on this page.
 
3 The other categories, Commentary, MP Corner & Advocacy no longer appear in the space under See All. Could you restore these?

lizconlan:
Resolves #1571 

* Alters the display order of blog posts on the home page to read left to right then down (as opposed to down then across like a newspaper)
* Allows 2 (or more) categories to be displayed on the blog categories page so that the main "See All" link can be to posts about That Week in Parliament and Implementations
* Fixes a bug inadvertently introduced in 7db85cb where a second Category model is imported, causing the code at the bottom of the home page which renders the top post in the 3 chosen blog categories to silently fail


lizconlan:
I've also reinstated the category summary text which was removed by mistake. It's only rendered if there is a single category though, otherwise the text makes no sense.

PR #1619, opened 2015-02-16
ZA: Reduce queries used on the organisation detail page

### LINKED ISSUE #1610: ZA: party pages are very slow to load
The ANC page stands out in Google Analytics as having a huge load time:

* http://www.pa.org.za/organisation/anc/

This should be a simple matter of using select_related / prefetch_related to reduce the number of queries made.

mhl:
Despite the careful use of select_related and prefetch_related in
pombola.core.views:OrganisationDetailView, the loading of an
organization with even a small number of people in it generate a huge
number of queries.

This is caused by the additional information requested by the South
African override of core/person_list_item.html, which does two costly
operations:

 1. Finding places associated with that person by iterating over
    position_set.all.current_unique_places and then accessing
    parent_place for each of those too.

 2. Calling the 'parties' method on each person, which creates a new
    queryset for each person.

Neither of these are all that easy to make more efficient while still
providing the same information without restructuring a lot of the
template code, so instead this commit just paginates the results so that
the page load times don't become excessive. (This reduces the number of
queries for a page load from thousands to about 80; still much higher
than we'd like, but the performance is OK.)

Fixes #1610

### DIFF
--- a/pombola/south_africa/templates/core/organisation_detail.html
+++ b/pombola/south_africa/templates/core/organisation_detail.html
@@ -39,6 +39,20 @@
     {% include "core/_position_listing.html" with positions=positions %}
   {% endif %}
 
+  <div class="pagination">
+    {% if positions.has_previous %}
+      <a href="?page={{ positions.previous_page_number }}">Previous</a>
+    {% endif %}
+
+    <span class="current">
+      Page {{ positions.number }} of {{ positions.paginator.num_pages }}
+    </span>
+
+     {% if positions.has_next %}
+      <a href="?page={{ positions.next_page_number }}">Next</a>
+    {% endif %}
+  </div>
+
 {% endblock %}
 
 {# Put a map in the profile_pic block #}
--- a/pombola/south_africa/views.py
+++ b/pombola/south_africa/views.py
@@ -325,6 +325,15 @@ class SAOrganisationDetailView(OrganisationDetailView):
             key=key_position_sort_last_name,
         )
 
+        paginator = Paginator(context['positions'], 20)
+        page = self.request.GET.get('page')
+        try:
+            context['positions'] = paginator.page(page)
+        except PageNotAnInteger:
+            context['positions'] = paginator.page(1)
+        except EmptyPage:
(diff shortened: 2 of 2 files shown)


pombola/south_africa/templates/core/organisation_detail.html
```diff
@@ -39,6 +39,22 @@
     {% include "core/_position_listing.html" with positions=positions %}
   {% endif %}
 
+  <div class="pagination">
+    <span class="step-links">
```
lizconlan: Having this extra `<span>` block inside the pagination container breaks the CSS button highlighting so the the whole pagination block rather than the target button lights up red on mouseover. Taking this out seems to fix it.

pombola/south_africa/templates/core/organisation_detail.html
```diff
@@ -39,6 +39,22 @@
...
+      {% if positions.has_previous %}
+        <a href="?page={{ positions.previous_page_number }}">previous</a>
+      {% endif %}
+
+      <span class="current">
```
lizconlan: Similarly wrapping the current page info inside a `<span>` tag here makes it look like a button. Taking it out makes it appear as text on the page instead, in which case consider removing the full stop as it could look out of place and lazily adding some padding with `&nbsp;` at either end.

lizconlan:
In a perfect world it would look more similar to the style used elsewhere, but especially as that looks non-trivial to implement, I think the simplified style's perfectly acceptable.

It has a few display bugs in it though, which make it behave a bit oddly, which I've commented on inline.

Definitely faster though!

pombola/south_africa/templates/core/organisation_detail.html
mhl: I've taken out the step-links span and I think it looks fine (I quite like the button look - the mouse pointer was in the "previous" link which is why it's highlighted):

pagination-2

Do you think that'll do? Otherwise, not having the button styling, but em-dashes after previous or before next (removing the full stop) looked fine too

lizconlan:
:+1:

PR #1625, opened 2015-02-19
Mzalendo: add date sort options to hansard search

### LINKED ISSUE #1607: Mzalendo - add date sorting to Hansard Search Results
Enable search results for Hansard - http://info.mzalendo.com/search/hansard/ - to be sortable in ascending / descending date order

lizconlan:
Fixes #1607 

* Adds a new field to `HansardEntryIndex` so that `start_date` is available to the sort function.
* Adds a drop-down box to the Hansard search form offering Sort by 'Relevance' (default), 'Oldest first' or 'Newest first'

I traded the `haystack.views` based `SearchView` for `HansardSearchView` which is a cut-down version of `SearchBaseView` minus sections and template_name resolution, mostly because trying to add `sort_by` to `SearchView` was giving me a headache. On the plus side, there is now scope to standardise the look of the Hansard search by including the existing partial templates for result items and pagination but I've avoided this in the first instance in case it introduces unwanted UI changes.

### DIFF
--- a/pombola/search/search_indexes.py
+++ b/pombola/search/search_indexes.py
@@ -48,8 +48,17 @@ if settings.ENABLED_FEATURES['hansard']:
     from pombola.hansard import models as hansard_models
 
     class HansardEntryIndex(BaseIndex, indexes.Indexable):
+        start_date = indexes.DateTimeField(null=True)
+
         def get_model(self):
             return hansard_models.Entry
+
+        def index_queryset(self, using=None):
+            """Used when the entire index for model is updated."""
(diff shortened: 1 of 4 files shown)


mhl:
I think that in the longer run, we'll want to switch Mzalendo to having a single search page with "advanced search" options, like People's Assembly: http://www.pa.org.za/search/

... but this is definitely a big improvement, and necessary foundation for that work anyway (the search index fix). :+1:

mhl:
@lizconlan Actually, there's one thing that needs to be fixed still - if you select newest-first ordering, then go to the second page, the ordering is lost

PR #1626, opened 2015-02-20
Add API help page for Mzalendo

lizconlan:
Overrides the core template to suppress the references to daily data dumps and "a number of ways of accessing ... data" if `dump_base_path` is not included in the view context

Note to self/deployer - relies on setting a valid `POPIT_API_URL` in the site config


mhl:
I think that overriding the `dump_base_path` in the context isn't really necessary - the data dumps it should always be there (and are now on pluto) in the known location, which would make this all a lot simpler, e.g. you wouldn't have to override the api help view specifically for Kenya.

mhl:
After squashing this down (which I'd suggest doing and force-pushing at this stage to make it clearer) the changes in pombola/kenya/urls.py aren't needed any more, so this can be an even smaller change :)

PR #1632, opened 2015-03-02
Add link asking Mzalendo users to complete research survey

jacksonj04:
* Solves https://github.com/mysociety/research/issues/50

### DIFF
--- a/pombola/kenya/static/js/survey.js
+++ b/pombola/kenya/static/js/survey.js
@@ -1,0 +1,72 @@
+(function() {
+
+    // Percentage of users to display the survey for
+    var percentage = 0.6,
+        cookie_time = null,
+        cookie_array = document.cookie.split(';'),
+        i, cookie, time, site_time, link, data, query_string;
+
+    for(i=0;i < cookie_array.length;i++) {
+        cookie = cookie_array[i];
+        cookie = cookie.replace(/^\s+/, '')
+        if (cookie.indexOf('ms_srv_t=') == 0) {
+            // Cookie for time...
+            cookie_time = cookie.substring('ms_srv_t='.length,cookie.length);
+        }
+        if (cookie.indexOf('ms_srv_r=') == 0) {
+            // Cookie for referrer...
+            cookie_referrer = cookie.substring('ms_srv_r='.length,cookie.length);
+        }
+    }
+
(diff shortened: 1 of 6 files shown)


pombola/kenya/static/js/survey.js
```diff
@@ -0,0 +1,75 @@
+(function() {
+
+    // Percentage of users to display the survey for
+    var p = 0.6;
+
+    // Is cookie existing?
+
+    var ct = null;
+    var ca = document.cookie.split(';');
+    for(var i=0;i < ca.length;i++) {
```
mhl: Could you explicitly hoist the variables in this function? I'm with Douglas Crockford on this - putting a var in the initializer of a for loop when it's really at function scope is confusing. (Similarly, the `var ct` below is unnecessary.)

pombola/kenya/static/js/survey.js
```diff
@@ -0,0 +1,75 @@
...
+        var c = ca[i];
+        while (c.charAt(0)==' ') {
```
mhl: Is there a particular reason to strip leading whitespace like this, rather than, say, `c.replace(/^\s+/, '')` ?

pombola/kenya/static/js/survey.js
```diff
@@ -0,0 +1,75 @@
...
+        } else {
+            // Not chosen to survey
+            // Set cookie to X
+            document.cookie = 'ms_srv_t=X; path=/';
+            ct = 'X'
+        }
+    }
+
+    // Only bother to do this if the cookie is set and the link exists
+    if (ct != 'X' && !! document.getElementById('ms_srv_wrapper')) {
```
mhl: The `!!` isn't needed here, I think - I don't see why this needs to be a boolean type as well as just truthy?

pombola/kenya/static/js/survey.js
```diff
@@ -0,0 +1,75 @@
+(function() {
```
mhl: Since this is a fairly long function, I think it'd be helpful to make the local variable names more descriptive than these one or two character names.

mhl:
I've made a few comments on the Javascript, but otherwise this looks fine.

mhl:
:+1:

Issue #1594, opened 2015-01-19
migrate to using django-popolo instead of pombola's core models


mhl:
I'm creating this new ticket to more accurately represent the work involved in this. (We had been using https://github.com/mysociety/pombola/issues/1551 as a placeholder for this job.) This plan probably needs some work.

The great advantage in doing this is that SayIt, when embedded in Pombola, and the rest of Pombola's code would use to the same Person objects. This would remove complex layers of indirection and state, and be a good step in reducing the complexity of maintaining the Pombola ZA scrapers. It's also a good move in terms of code reuse and moving Pombola to be more like a collection of components.

- [ ] Subclass the django-popolo models (multi-table inheritance) and add any missing fields which were in the Pombola core models.
- [ ] Write migrations to migrate data from the core Pombola models to the new django-popolo models.
- [ ] Make any use of the original Pombola core models an error (e.g. by replacing the Person class with `class Person: pass`
- [ ] Update all code that uses the original models until the tests pass
- [ ] Update all the rest of the code that uses the original models but isn't covered by tests (adding tests at the time if possible)
- [ ] Update Pombola ZA's SayIt integration to eliminate the SASpeakerRedirectView and the linking of people between saying and Pombola via PopIt IDs (see https://github.com/mysociety/pombola/issues/1551 )
- [ ] :cocktail: :wine_glass: :fireworks:

davea:
A summary of the progress so far (in [1594-django-popolo-migration](https://github.com/mysociety/pombola/tree/1594-django-popolo-migration)):

 * Fields required by Pombola and not provided by django-popolo have been identified
 * Subclasses of `popolo.models.{Organization,Person,Post}` have been created as `pombola.core.Popolo{Organization,Person,Post}` with the missing fields added.
 * `core.Organisation.name` needed 200 chars but `popolo.Organization.name` is defined with 128 `max_length`, and you can't redefine fields in model subclasses so we're using [our own fork](https://github.com/mysociety/django-popolo/tree/increase-charfield-lengths) of django-popolo with `max_length` set to 4000 for a few fields.
 * A WIP [migration](https://github.com/mysociety/pombola/blob/1594-django-popolo-migration/pombola/core/migrations/0062_migrate_core_models_to_popolo.py) has been created that so far migrates the 'simple' fields for Organisations along with the Contact relations (which are migrated to popolo's `ContactDetail` model).

[...]

PR #1633, opened 2015-03-03
ZA: Paginate Hansards page

lizconlan:
Adds pagination controls to the Hansard and Committee Minutes pages (they're derived from the same Index view).

Formulates the Sections queries in a top-down manner rather than the previous drill-up strategy. No longer retrieves individual Speech entries as only the speech counts are used by the display template.

Force groups the subsections by named section in SQL to remove duplicate section names as this was causing uneven pagination effects where one page could be 3 sections or more longer than expected while the next page may only consist of 2 sections. As this strategy has a knock-on effect of occasionally creating a really long list of subsections - which results in a large number of database lookups to render the links - the page length for Hansard listings has been shortened from 25 to 15 sections; hopefully this shouldn't be a major problem as adding pagination reduces the need to display as many things as possible on the first page. The Committee Minutes page is largely unaffected so remains a 25 section per page view.

### DIFF
--- a/pombola/south_africa/fallback_urls.py
+++ b/pombola/south_africa/fallback_urls.py
@@ -34,9 +34,9 @@ urlpatterns = [
 # Make sure the top level custom indexes work:
 
 urlpatterns += patterns('',
-    url(r'^hansard/$', SAHansardIndex.as_view(), name='section-list-hansard'),
-    url(r'^committee-minutes/$', SACommitteeIndex.as_view(), name='section-list-committee-minutes'),
-    url(r'^question/$', SAQuestionIndex.as_view(), name='section-list-question'),
+    url(r'^hansard/?$', SAHansardIndex.as_view(), name='section-list-hansard'),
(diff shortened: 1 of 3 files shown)


dracos:
It sounds like it might be nice to get pombola's SayIt upgraded (it has e.g. start date and more metadata on sections, and blank section title display), which would make this much simpler :)

mhl:
@dracos Yep, indeed, we've talked about that but it's quite a big job (since it only makes sense to do the SayIt upgrade after #1594) and there is prioritization brokenness involved :(

mhl:
As we discussed in IRC, one problem (that might as well go into this pull request, although it was a problem that existed before) is to change the match for `/hansard/`, `/committee-minutes/` etc. in `pombola/south_africa/fallback_urls.py` to make the `/` optional, otherwise you get the default section view for the top level.

mhl:
:+1: Good to deploy now - thanks!

lizconlan:
:smile_cat: thanks, will get it merged in

PR #1638, opened 2015-03-09
ZA: update person profile page

### LINKED ISSUE #1602: ZA: Updates to person profile pages (inactive people, date of death, contact details)
Profile pages need updating:

- [ ] For people with one or less positions (one being just a party membership), former positions should be listed on the default tab. This would avoid otherwise blank profiles like [this](http://www.pa.org.za/person/suzanne-marylan-van-aardt/) and provide information for people who are no longer MPs ([e.g.](http://www.pa.org.za/person/max-vuyisile-sisulu/)) or who have passed away ([e.g.](http://www.pa.org.za/person/mario-gaspare-oriani-ambrosini/)).

 This should possibly be limited to certain positions (legislature, executive, election lists, etc.) - as some MPs have a large number of former positions ([e.g.](http://www.pa.org.za/person/mario-gaspare-oriani-ambrosini/)). 

- [ ] If a person has passed away and their membership of their party has therefore ended, their affiliation should still be displayed in their profile in the sidebar ([e.g.](http://www.pa.org.za/person/mario-gaspare-oriani-ambrosini/)).

- [ ] Dates of death should be displayed.

- [x] Addresses (physical, postal) should be displayed on profiles if entered.

lizconlan:
Fixes #1602

* Inserts date of death at the top of the summary box, if applicable
* Retains the person's party affiliation if the deceased's membership has ended
* If there is only one current position (this is lazily assumed to be a party membership) - or nothing at all - to display on the profile summary, positions within "important organisations" + Electoral List positions will appear under the heading "Former Positions" (this should exclude party and committee memberships to avoid duplicating the contents of the Experience tab, especially where there is a long list of former positions)

### DIFF
--- a/pombola/south_africa/templates/core/person_detail.html
+++ b/pombola/south_africa/templates/core/person_detail.html
@@ -108,6 +108,9 @@
   <div id="profile" class="tab-content tab-active">
 
     <div class="person-summary">
+      {% if person.date_of_death %}
+        <p>Died {{ person.date_of_death }}</p>
+      {% endif %}
       {{ person.summary }}
     </div>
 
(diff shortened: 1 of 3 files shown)


mhl:
:+1: A few small comments, but otherwise this looks good!

PR #1639, opened 2015-03-11
Convert install doc to markdown

stevenday:
Install docs were a simple text file, while everything else is markdown, this PR changes that, so it looks nicer.


mhl:
:+1: Thanks, @stevenday

PR #1641, opened 2015-03-20
ZA: Replace election list org links with links to election candidate lists

### LINKED ISSUE #1603: ZA: Viewing election list organisations should always display the correct view
If one clicks on an election list on a person's profile (e.g. [this profile](http://www.pa.org.za/person/suzanne-marylan-van-aardt/) under experience), the [standard organisation page](http://www.pa.org.za/organisation/icosa-provincial-western-cape-election-list-2014/) is displayed. This page has no people listed as positions have been ended. The link should direct to the [relevant election page](http://www.pa.org.za/election/2014/national/western-cape/icosa/) instead.

lizconlan:
On profile pages, the link to e.g. "3rd Candidate at Inkatha Freedom
Party National Election List 2014 (Election List)" were resolving to
blank pages as an Election List is only valid for a short time and will
have no associated members once the election is complete. This change
replaces these links (except in the case of a "Regional" election link
which look as though it may be a duplicate of "Provincial" election data)
with a link to the appropriate election candidate list instead.

Fixes #1603

### DIFF
--- a/pombola/south_africa/templatetags/za_election_list_links.py
+++ b/pombola/south_africa/templatetags/za_election_list_links.py
@@ -1,0 +1,38 @@
+from django import template
+import re
+
+register = template.Library()
+
+@register.filter()
+def get_party_slug(org_slug):
+    """Takes the original slug of an election list 'organisation' and returns
+    the party slug
+    """
+
+    re_matches = decompose_slug(org_slug)
+    return re_matches.group('party')
+
+@register.filter()
+def get_place_slug(org_slug):
+    """Takes the original slug of an election list 'organisation'
+    and returns the place (province) slug
+    (or None if the slug contains no province information)
+    """
+
+    re_matches = decompose_slug(org_slug)
+    return re_matches.group('place')
+
+def decompose_slug(slug):
+    """Returns regexp matches
+    """
+    election_list_slug_re = re.compile(
+        r'''^
+        (?P<party>.*)-
+        (?P<place_type>provincial|national|regional)-
+        (?:(?P<place>.*)-)*
+        election-list-
+        (?P<year>.*)
+        $''',
+        re.VERBOSE
+    )
+    return election_list_slug_re.search(slug)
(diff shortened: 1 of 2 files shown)


pombola/south_africa/templatetags/za_election_list_links.py
```diff
@@ -0,0 +1,38 @@
...
+def get_place_slug(org_slug):
+    """Takes the original slug of an election list 'organisation' and returns
+    the place (province) slug
+    """
+
+    stripped = strip_slug(org_slug)
+    parts = stripped.split("--")
+    return parts[1]
+
+def strip_slug(slug):
```
mhl: I think I'd find this easier to have confidence in if it was picked apart with a regular expression, e.g. in this module you could add:

```python
election_list_slug_re = re.compile(
    r'''^
    (?P<party>.*)-
    (?P<place_type>provincial|national)-
    (?P<place>.*)-
    election-list-
    (?P<year>.*)
    $''',
    re.VERBOSE
)
```

... which seems like a clearer way of decomposing the slug. Then you can do:

```python
m = election_list_re.search(org_slug)
```

... and either raise an exception if `m` is `None` or return `m.groupdict()` otherwise. The dict will make the other template filter functions read better as well.

lizconlan: Hope this is ok, I've reworked the regular expression to account for both known slug formats:

```python
election_list_slug_re = re.compile(
    r'''^
    (?P<party>.*)-
    (?P<place_type>provincial|national)-
    (?:(?P<place>.*)-)*
    election-list-
    (?P<year>.*)
    $''',
    re.VERBOSE
)
```

examples:
  `ifp-provincial-western-cape-election-list-2014`
   `ifp-national-election-list-2014`

lizconlan:
TODO: fix for regional elections (redirect to 'provincial'?)

mhl:
:+1:

PR #1642, opened 2015-03-24
KE: Update Hansard scraper to use new site structure

### LINKED ISSUE #1634: Urgent Hansard update
Dear @paullenz and @mhl,

I noticed Mzalendo's hansard is not automatically syncing with the Parliament site. 
Do you need to update the code? During the Christmas break, the Kenyan Parliament revamped its website.

Please advise.

Regards,

Jessica

lizconlan:
The web scraper in the `hansard_check_for_new_sources` management
command has not been finding new Hansard PDF files since the Kenyan
Parliament site was relaunched over the Christmas 2014 break. This
update adjusts the existing scraper for the new URLs and HTML layout.

As the National Assembly listings page now only links to the last 15
documents, code has been added to traverse across multiple pages if the
final document on the current page is treated as a new source. The hope
is that this should be reached very infrequently after the initial
catchup scraper run.

Resolves #1634

### DIFF
--- a/pombola/hansard/management/commands/hansard_check_for_new_sources.py
+++ b/pombola/hansard/management/commands/hansard_check_for_new_sources.py
@@ -1,9 +1,17 @@
 
-# This script changed extensively when the Kenyan Parliament website changed after the 2013 Election.
+# This script was first changed extensively when the Kenyan Parliament website
+# changed after the 2013 Election.
+#
+# The pre-2013 Election version can be seen at:
+#
+#    https://github.com/mysociety/pombola/blob/7181e30519b140229e3817786e4a7440ac08288d/mzalendo/hansard/management/commands/hansard_check_for_new_sources.py
+#
+# It was then changed again after the Kenyan Parliament website was
+# given an overhaul during the Christmas 2014 break.
 #
 # The previous version can be seen at:
 #
-#    https://github.com/mysociety/pombola/blob/7181e30519b140229e3817786e4a7440ac08288d/mzalendo/hansard/management/commands/hansard_check_for_new_sources.py
+#    https://github.com/mysociety/pombola/blob/ec4a44f7d7e0743426aff87b59e4bfa54250ec1c/pombola/hansard/management/commands/hansard_check_for_new_sources.py
 
 import pprint
 import httplib2
@@ -12,6 +20,7 @@ import datetime
 import sys
 import parsedatetime as pdt
 from warnings import warn
+from urlparse import urlsplit, urlunsplit
 
 from BeautifulSoup import BeautifulSoup, BeautifulStoneSoup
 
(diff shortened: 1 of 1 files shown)


pombola/hansard/management/commands/hansard_check_for_new_sources.py
```diff
@@ -66,9 +72,16 @@ def process_url(self, list_page, url, verbose):
             convertEntities=BeautifulStoneSoup.HTML_ENTITIES
         )
 
-        spans = soup.findAll( 'span', 'contenttype-repositoryitem summary')
+        base_url = '/'.join(url.split('/')[:3])
```
mhl: I found this a bit unclear - it's probably simpler to use [the urlparse module](https://docs.python.org/2/library/urlparse.html) for this, I think, e.g.

```python
from urlparse import urlsplit, urlunsplit
url_parts = urlsplit(url)
base_url = urlunsplit((url_parts.scheme, url_parts.netloc, '', '', ''))
```

pombola/hansard/management/commands/hansard_check_for_new_sources.py
```diff
@@ -95,42 +108,67 @@ def process_url(self, list_page, url, verbose):
...
+
+        # Scrape the next page if the last item on the page was added
+        # as a new source and the next_page_button contains a link
+        # (assumes that Hansard docs are being listed - and consumed -
+        # in date order, newest first)
+        #
+        # Hopefully requiring a second page of data will be a rarity;
+        # potentially fragile if the connection is lost before traversing
+        # all necessary pages
+        if get_next_page:
```
mhl: Rather than have some of the next page logic here, and making a recursive call, I'd be tempted to write a generator, `get_index_pages`, say, that takes a URL like `http://www.parliament.go.ke/index.php/the-senate/house-business/hansard` and `yield`s a soup and base_url for each index page, so that all the page iteration logic is in a separate function (except for when you decided to break out of the iteration). It'd be fine like this too though.

pombola/hansard/management/commands/hansard_check_for_new_sources.py
```diff
@@ -66,9 +72,16 @@ def process_url(self, list_page, url, verbose):
...
+
+        next_url = self.get_next_url(soup, base_url)
+        get_next_page = False
 
-        links = [ span.a for span in spans ]
+        # Grab each of the blocks on the current page which are supposed to
+        # contain a link to a PDF file.
+        link_sections = soup.findAll( 'div', 'catItemView groupPrimary' )
+
+        links = [ section.find( 'div', 'catItemAttachmentsBlock' ).a for section in link_sections ]
```
mhl: Some minor PEP 8 violations here (which the code you're replacing had too, but I think it's worth changing) - there shouldn't be a space after an opening `[` or `(` or before a closing `]` or `)`, and a few of these lines are over 80 characters; I'd probably split this one as:

```
links = [
    section.find( 'div', 'catItemAttachmentsBlock' ).a
    for section in link_sections
]
```

PR #1646, opened 2015-03-30
ZA: Add link prompting ZA users to take survey

jacksonj04:
Similar to #1632, just with a different country

Part of https://github.com/mysociety/research/issues/4


mhl:
Since these two copies of `survey.js` are identical:

```
./pombola/south_africa/static/js/survey.js
./pombola/kenya/static/js/survey.js
```

... you could put that into static files in core, or even in `commonlib/jslib/` so it can be used on other sites too.

The survey URL could be setting in `general.yml` as well, and just wrap the `.ms_srv_wrapper` div in a conditional that only includes it if that's set, dunno if that'll be worth it though, I guess it depends how often you're going to want to run surveys on Pombola sites.

jacksonj04:
The logic for keeping two distinct copies of the JS was that it allows the surveys to be removed independently by a single git revert, rather than needing to re-extract the functionality.

mhl:
You could make the Javascript only run if the `div.ms_srv_wrapper` is present, and if you did the second part I suggested, enabling it per site would just be a matter of changing the survey URL config setting to a URL or the empty string. (I don't mind about having the Javascript present even if it's not used, it'll be tiny minified.)

jacksonj04:
I've made the JS now exist centrally, and only run if the link element is present. Each survey link, however, has its own URL, link text and display percentage, so I'm not convinced it's easier to add these as config variables rather than allow each site to independently choose where to show the link (by adding it to templates), and thus just put the 'config' in there.

mhl:
:+1:

PR #1647, opened 2015-03-31
Mzalendo: allow Hansard parser to deal with Joint Sittings

### LINKED ISSUE #1645: Mzalendo - better Hansard scraper recognition of venues
Should allow for Joint Sittings - currently doesn't pick up the venue or the start time. The start time is a bit tricky as for joint sittings it is written as "Parliament met at fifty five minutes past three o'clock" rather than the expected "The House met at 3.55 p.m."

broken out from #1634

lizconlan:
Adds a new reg exp for Joint Sitting headings and attempts to parse the start time from the accompanying "fifty five minutes past Two o'clock" style format.

Have also made a few tweaks to the reg exps for Senate and National Assembly sittings which should help to scoop up a few cases where Venue wasn't being picked up properly having notice that Senate sittings can also be written as "The House met at 2.30 p.m. in the Senate Chambers" (as opposed to the expected string "The Senate met....".

Fixes #1645

### DIFF
--- a/pombola/hansard/kenya_parser.py
+++ b/pombola/hansard/kenya_parser.py
@@ -18,14 +18,14 @@ class KenyaParserCouldNotParseTimeString(Exception):
     pass
 
 
-class KenyaParser():  
-    
+class KenyaParser():
+
     @classmethod
     def convert_pdf_to_html(cls, pdf_file):
         """Given a PDF parse it and return the HTML string representing it"""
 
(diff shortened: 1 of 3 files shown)


lizconlan:
Sorry for trashing the conversation but I couldn't work out how to do the `fixup`s cleanly :cry:

mhl:
:+1:

PR #1648, opened 2015-04-13
Make the survey / experiment views easier to reuse

mhl:
The views used for the County Performance experiments could be
reused for other similar pages, but several properties of those
experiments were hardcoded in the views. In addition, it makes
sense for the generic view code to be in the experiments application
instead of kenya.


jacksonj04:
Looks :+1: to me.

PR #1667, opened 2015-06-09
Kenya Experiment Page

### LINKED ISSUE #1644: Implement the next round of MIT research pages for Mzalendo
Ideally would like this done by the 10th April

jacksonj04:
Sets up a new Experiment Page for Mzalendo around the Youth Employment Bill.

Restructures the original Experiment code a little so that it can be reused here along the way.

Closes #1644


lizconlan:
hmm 9cff063 / cd07fb2 doesn't seem to work in Opera Mini :crying_cat_face:

pombola/kenya/views.py
```diff
@@ -45,6 +50,19 @@
...
             'agroup': ('under', 'over'),
         },
+    },
+    'youth-employment-bill': {
+        'session_key_prefix': 'MIT3',
+        'base_view_name': 'youth-employment',
+        'pageview_label': 'youth-employment',
+        'template_prefix': 'youth',
+        'experiment_key': settings.YOUTH_EMPLOYMENT_BILL_EXPERIMENT_KEY,
+        'qualtrics_sid': 'SV_5hhE4mOfYG1eaOh',
```
mhl: It's probably best to set this to `None` until we have the real new survey ID, just so we don't forget to update it. (Or was it clear that they'd be using the same survey for this experiment too?)

mhl:
This seems to generally work well; I've checked and it looks like events are being recorded for all the actions. The other things that struck me are:

* It would be good to make the voting lower friction (keeping people on the same page) by Javascript enhancement.

* I think we might be going a bit far in using the Feedback mechanism to store arbitrary data; it sort of makes sense for free-form comments (since that's the type of data already being stored by the feedback application, and at one stage we thought that Mzalendo admins might be going through them, manually forwarding them). However, for things like the name of the constituency submitted and the yes / no value that would seem to be better stored as attributes in the extra_data of the event.

* I'm not sure if we heard back from Leah about whether they need time-on-page tracking, which we didn't do previously - that would be worth checking.

* The event metadata seems inconsistent between:
```
       category: mp-cares   action: click       label: yes
       category: form       action: submit-no   label: support
```
   I think in the former case, this would be more consistent
   with the other events:

       category: form       action: click-yes   label:mp-cares

pombola/kenya/views.py
lizconlan: Oops, lazy copy and paste. Will change it to None to avoid accidentally serving the wrong survey

mhl:
Notes from Liz and I going through the code together on things still to do:

* Set up the GA experiment
* Insert the real link to the bill
* Switch the "support" view to store answer in extra_data, not feedback
* Make consistent the click-yes vs submit actions for "cares" and "supports"

* Should riveted be told to stop pinging the server after a certain time, so we don't use up people's data
* Remove some repeated code from urls.py
* Javascript form validation
* Make sure AJAX request that cause events to be stored are POST not GET (mp-cares)
* AJAX request should show "in progress" and then completion message

PR #1692, opened 2015-07-01
1665 tag cloud

duncanparkes:
Make a wordcloud app and use it for Odekro.

I'm sure there are further bits here that could be made tidier, and there certainly could me more testing.


wordcloud/wordcloud.py
```diff
@@ -0,0 +1,46 @@
...
+from pombola.hansard import models as hansard_models
+
+
+BASEDIR = os.path.dirname(__file__)
+# normal english stop words and hansard-centric words to ignore
+STOP_WORDS = open(os.path.join(BASEDIR, 'stopwords.txt'), 'rU').read().splitlines()
+
+
+def recent_entries(max_entries=20):
+    return SearchQuerySet().models(hansard_models.Entry).order_by('-sitting_start_date')[:max_entries]
```
mhl: The highest level comment I have on this PR is probably "why use Haystack / Elasticseach for this rather than just the ORM for Entry?" (There may well be a good reason, I just can't think of it.)

wordcloud/wordcloud.py
```diff
@@ -0,0 +1,46 @@
...
+from operator import itemgetter
+import re
+
+from haystack.query import SearchQuerySet
+from pombola.hansard import models as hansard_models
+
+
+BASEDIR = os.path.dirname(__file__)
+# normal english stop words and hansard-centric words to ignore
+STOP_WORDS = open(os.path.join(BASEDIR, 'stopwords.txt'), 'rU').read().splitlines()
```
mhl: I think STOP_WORDS should be a set - since the code's only ever checking membership it'll be way faster. Also, I'd write this as:

```python
with open(os.path.join(BASEDIR, 'stopwords.txt'), 'rU') as f:
    STOP_WORDS = set(f.read().splitlines())
```

... for the nice file-closing-on-exceptions behaviour.

wordcloud/urls.py
```diff
@@ -0,0 +1,9 @@
+from django.conf.urls import patterns, url
+
+from .views import wordcloud
+
+urlpatterns = patterns(
+    '',
+    url(r'^wordcloud/$', wordcloud, name='wordcloud'),
```
mhl: The renaming to wordcloud makes sense, but I'd keep `tagcloud` as an alias in `urls.py` just for the transition - we can remove it later.

wordcloud/management/commands/generate_wordcloud.py
```diff
@@ -0,0 +1,37 @@
...
+
+    def handle(self, *args, **options):
+        if len(args) > 1:
+            raise CommandError
+
+        cache_dir = settings.WORDCLOUD_CACHE_DIR
+        try:
+            os.mkdir(cache_dir)
+        except OSError:
+            # Probably the directory exists already
```
mhl: You can check if exc.errno == errno.EEXIST - this code's already `pombola/kenya/management/commands/iebc_api.py` so it might be worth moving that definition of mkdir_p somewhere common.

wordcloud/views.py
```diff
@@ -0,0 +1,22 @@
+import os
+
+from django.http import HttpResponse
+from django.utils import simplejson
+from django.conf import settings
+
+from .wordcloud import popular_words
+
+
+def wordcloud(request, max_entries=30):
```
mhl: It might be worth adding a Cache-Control header to increase the time the wordcloud data is cached for (e.g. to a few hours)

wordcloud/management/commands/generate_wordcloud.py
```diff
@@ -0,0 +1,37 @@
...
+    'wordcloud.json'. If the file path is absolute it will just be used
+    as is.
+    """
+
+    help = 'Generate wordcloud data'
+    args = '<file>'
+
+    def handle(self, *args, **options):
+        if len(args) > 1:
+            raise CommandError
```
mhl: It's be nice if this CommandError had a usage message as its message

wordcloud/views.py
```diff
@@ -0,0 +1,22 @@
...
+    """ Return tag cloud JSON results"""
+    cache_path = settings.WORDCLOUD_CACHE_PATH
+    if os.path.exists(cache_path):
+        with open(cache_path) as cached_file:
+            content = cached_file.read()
```
mhl: If there's some long caching set on this (and Ian's using Varnish or some other caching reverse proxy) it may not matter, but this looks like a case for using x-sendfile to serve the generated file more efficiently.

wordcloud/wordcloud.py
```diff
@@ -0,0 +1,46 @@
...
+            cloudlist.append(
+                {
+                    "text": word,
+                    "weight": words.get(word),
+                    "link": "/search/hansard/?q=%s" % word,
+                    }
+                )
+
+        sortedlist = sorted(cloudlist, key=itemgetter('weight'), reverse=True)[:max_words]
+    except:
```
mhl: Bare `except`! :postal_horn: :) This will catch typos in variable names, etc. as well - shouldn't any exception be propagated from here anyway?

wordcloud/wordcloud.py
duncanparkes: I copied this from the original code, and as I did it I had the same thought as you. The reasons I kept the ElasticSearch version was that we can use the all the content together thing that's used by search for the words - so whatever model, we can have all the words from all the wordy bits, as it were. (I realize we'd have to refactor a tiny bit to use a different model, but you know what I mean).

wordcloud/urls.py
duncanparkes: I'll add this to pombola/urls.py too.

wordcloud/wordcloud.py
duncanparkes: Agreed!

wordcloud/management/commands/generate_wordcloud.py
duncanparkes: I'll just copy it for now - I'm not sure where that somewhere common should be!

wordcloud/views.py
duncanparkes: I've added some caching for now.

wordcloud/wordcloud.py
duncanparkes: I think I just copied that from the old code. I can't see what it could catch that would be legit, let's just take it out.

duncanparkes: I can't see any good excuse for using `sorted` rather than sorting in place either.

duncanparkes:
@mhl I think I've made all the changes you suggested here. If you could sanity check them that would be great - and in particular, does 9dec103 look good to you? I'm not sure how to test that one without a proper web server.

wordcloud/management/commands/generate_wordcloud.py
```diff
@@ -0,0 +1,40 @@
...
+    def handle(self, *args, **options):
+        if len(args) > 1:
+            raise CommandError("Usage: python manage.py generate_wordcloud [filepath]")
+
+        cache_dir = settings.WORDCLOUD_CACHE_DIR
+        try:
+            os.mkdir(cache_dir)
+        except OSError as err:
+            # Probably the directory exists already
+            if exc.errno == errno.EEXIST and os.path.isdir(path):
```
mhl: There's a missing import for `errno`

duncanparkes: Good point! Probably could have run flake8 on that file too.

wordcloud/views.py
```diff
@@ -0,0 +1,26 @@
...
+
+
+@cache_page(60*60*4)
+def wordcloud(request, max_entries=30):
+    """ Return tag cloud JSON results"""
+    cache_path = settings.WORDCLOUD_CACHE_PATH
+    if os.path.exists(cache_path):
+        response = HttpResponse()
+        response['Content-Type'] = 'application/json'
+        response['X-Sendfile'] = cache_path.encode('utf-8')
```
mhl: I think this will also need `libapache2-mod-xsendfile` added to `conf/packages` and you'll need to add `XSendFile        On` and set `XSendFilePath` in the Apache configuration. For trying it out, if you don't have a local Apache, you could create a dev vhost on firefly or do it in a Vagrant box.

I think I mentioned this before, but I think I'd put this generated file in a subdirectory under `MEDIA_ROOT`.

PR #1693, opened 2015-07-01
Stop using Google Analytics' Content Experiments for randomizing pages

mhl:
All the data that MIT want is being stored in our database, and we're
not using Google Analytics for anything except picking a random variant
of the page.  Using it for randomization also has a few difficulties:
    
 * It does a redirect after the page load, and sometimes we don't seem
   to be getting that redirect (because of Javascript failures or
   network problems, we assume)
    
 * There's not a natural goal we can set for the content experiment, so
   we've just been setting 99.5% confidence that one variant increases
   page views across the site as a whole, which is unlikely to happen
   and stop the experiment, but still seems silly.
    
 * The code in the view is unnecessarily complicated to deal with
   essentially ignoring page views without a utm_expid.
    
So, this commit switches to picking a variant randomly ourselves, so
there should only be a single page load after clicking the Facebook
advertisement.

### DIFF
--- a/pombola/experiments/views.py
+++ b/pombola/experiments/views.py
@@ -38,6 +38,7 @@ class ExperimentViewDataMixin(object):
     experiment_key = None
     qualtrics_sid = None
     variants = None
+    default_variant = None
     demographic_keys = None
     major_partials = None
 
@@ -99,7 +100,7 @@ class ExperimentViewDataMixin(object):
             key='variant',
             parameters=parameters,
             allowed_values=self.variants,
-            default_value='n')
+            default_value=self.default_variant)
         for key, possible_values in self.demographic_keys.items():
             result[key] = sanitize_parameter(
                 key=key,
--- a/pombola/experiments/views.py
+++ b/pombola/experiments/views.py
@@ -45,6 +45,12 @@ class ExperimentViewDataMixin(object):
         prefix = self.session_key_prefix
         return prefix + ':' + key
 
+    def get_session_value(self, key):
+        return self.request.session.get(self.qualify_key(key))
+
+    def set_session_value(self, key, value):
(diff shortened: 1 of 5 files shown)


lizconlan:
Hmm, yes. While I think that using Google Analytics just to get a new variant string introduces an unhelpful extra step, I share a general sense of unease that we're breaking something but I'm not sure what that something is (double-checking? not sure). It's a shame that we can't asynchronously suggest a variant key to Google.

That said, I think that experimenting to see whether the drop-off rate improves without the extra redirect is well worth trying. Maybe we could use the lack of an `experiment_key` to not supply the Google Analytics code and choose our own variant key instead so that we can choose whether to use the Google stuff on a per-experiment basis (or is that going to be even more complicated for no good reason?)

mhl:
With regard to the drop-off rate, since the drop-off was between initial page load and GA-provoked redirect, and now we're not needing the latter, my assumption is that this can't not fix that problem! I guess my worry is that that doesn't help us to understand what was going wrong.

On your second point, I originally started writing it with a flag in the `EXPERIMENT_DATA` to make it option to use GA to pick the variant, but that made the view code much less clear (or my brain wasn't good enough to implement it in a way that was obviously correct)...

dracos:
Just to note for future some GA things that might be of interest: you can ask GA for a variant without a redirect and make changes with JS/CSS/what have you, it's what FMS A/B testing does, see https://developers.google.com/analytics/solutions/experiments-client-side

"It's a shame that we can't asynchronously suggest a variant key to Google." - you can use your own experiment IDs, map them to Google ones and send if you do want to do that, see https://developers.google.com/analytics/solutions/experiments-server-side

mhl:
Thanks, @dracos - I'll bear that in mind for the future.

PR #1697, opened 2015-07-07
Fix location search form layout

### LINKED ISSUE #1657: add search button to rep locator page
Currently user must hit enter to submit the search.
Add a submit button with "search" or somesuch text on it so user doesn't have to know the secret of pressing "enter".

lizconlan:
This adds a submit button to the location search form and applies the search form stylings that are used elsewhere.

It should also prevent a blank map from being displayed when there are no search results

(Applies to the base search/location template as well as the custom template for South Africa)

Fixes #1657

### DIFF
--- a/pombola/search/templates/search/location.html
+++ b/pombola/search/templates/search/location.html
@@ -6,7 +6,10 @@
 {% endblock %}
 
 {% block search_form %}
-  <input type="text" name="q" value="{{ query }}" id="id_q" />
+  <div class="inline-search-box">
+    <input type="text" name="q" value="{{ query }}" id="id_q" placeholder="Enter your location&hellip;"/>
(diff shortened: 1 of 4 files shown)


pombola/search/templates/search/location.html
```diff
@@ -6,35 +6,40 @@
...
-            name: "{{ result.address }}"
-          });
-        </script>
-      {% endfor %}
-    </ul>
-
-    <br style="clear: both;">
-
-  </div>
+  {% if geocoder_results %}
```
mhl: It would be nice to include an error message saying, "The location 'foobar' couldn't be found" or something like that when there are no results.

mhl:
A couple of small things to fix, but otherwise looks good :+1:

PR #1699, opened 2015-07-09
Fix the ShineYourEye PU number lookup

### LINKED ISSUE #1675: PU lookup not working on ShineYourEye
PL to get more details

lizconlan:
This was breaking because the regex that was checking for a valid
PU number was expecting a PU formatted string and we were passing
it the raw query instead, so the PU lookup was being bypassed
unless the search string took the form of OD:1:2.

Although we don't do anything with the last part of a full 4 part
PU number, we are now accepting the full 01:02:03:04 format as
input for our PU number lookup.

Fixes #1675

### DIFF
--- a/pombola/nigeria/tests.py
+++ b/pombola/nigeria/tests.py
@@ -1,11 +1,14 @@
 import unittest
 import doctest
+import re
 from . import views
 
 from django.test import TestCase
+from django_webtest import WebTest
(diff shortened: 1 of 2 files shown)


mhl:
:+1: :ice_cream: Thanks, looks good and it works locally for me

PR #1702, opened 2015-07-14
ZA: Add management command to generate adword csvs

### LINKED ISSUE #1683: csv files for adwords
Could someone provide me with some csv files for pa.org.za Adwords set-up, please?

Here's what I'd like, ideally:

a) Names of all representatives (common names if available, as I know some can be very long), paired with the URL of that rep's page. If it's not too difficult, it'd also be handy to have their positions, parties and which parliament they are in, but that's not as important)
b) Names of provinces, paired with the province page URLs
c) Names of parties, paired with the party page URLs

And maybe, again only if it's not too tricky:
d) The names of constituency offices, but in the forms people are likely to search for them, which presumably means taking the given name like "ANC Constituency Office (927): Odendaalsrus " and changing it to something like "Odendaalsrus Constituency Office". 

(@paullenz @JenMysoc I'm thinking about the talk you gave at the meet-up and specifically the part where some users said they weren't really interested in the concept of constituency offices. Is it worth running these proposed keyword groups past PA before I go ahead and set up the ads?)

lizconlan:
Writes out 4 csv files with data for representatives, provinces, parties and constituency offices.

The queries involved have less common ground with https://github.com/mysociety/pombola/blob/master/pombola/kenya/management/commands/kenya_adwords_csv.py than I expected (although I may well have missed the obvious here) so I have not attempted to pull out the common functionality into a base adwords_csv command.

Fixes #1683

### DIFF
--- a/pombola/south_africa/management/commands/south_africa_adwords_csv.py
+++ b/pombola/south_africa/management/commands/south_africa_adwords_csv.py
@@ -1,0 +1,142 @@
+# -*- encoding: utf-8 -*-
+
+import csv
+import os
+import re
+import sys
+from optparse import make_option
+
+from pombola.core.models import Organisation, Place
+from pombola.south_africa.models import ZAPlace
+from django.core.management.base import NoArgsCommand
+
+
+data_directory = os.path.join(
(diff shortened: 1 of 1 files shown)


mhl:
:+1:

Issue #1669, opened 2015-06-11
ShineYourEye - people not showing as party members


paullenz:
On page e.g. http://www.shineyoureye.org/person/theodore-ahamefule-orji/experience/ even though they are listed as an ongoing party member, on the LHS is says "not a member of any party"

dracos:
The LHS looks for orbs with kind slug "party", but the party in the main section has kind slug "political-party".

duncanparkes:
@dracos is correct. An organisation is considered to be a party for these purposes if it has the organisation kind with slug 'party'. Changing the slug of the OrganisationKind on shineyoureye looks to me as though it would change a single URL: `/organisation/is/political-party/` would become `/organisation/is/party/`. As far as I can see there is no reference to this URL from anywhere else in the codebase except for `/parties/` which is actually expecting the slug to be `party` and so is currently broken.

Let's change the slug of the 'Political Party' OrganisationKind to 'party' to match the expectations of the code.

mhl:
I'm happy with that so long as you add a redirect for `/organisation/is/political-party/` - you can try it out on nigeria.pombola.staging.mysociety.org

PR #1704, opened 2015-07-16
Allow slug redirects for OrganisationKinds.

duncanparkes:
Use SlugRedirectMixin to allow an old organisation kind slug to
redirect to a replacement slug.

If this change is accepted, #1669 just needs a row adding to the slug redirects table to redirect the  'political-party' `OrganisationKind` to 'party'.

### DIFF
--- a/pombola/core/models.py
+++ b/pombola/core/models.py
@@ -561,6 +561,10 @@ class OrganisationKind(ModelBase):
     class Meta:
        ordering = ["slug"]
 
+    @models.permalink
+    def get_absolute_url(self):
+        return ('organisation_kind', (self.slug,))
+
 
 class OrganisationQuerySet(models.query.GeoQuerySet):
     def parties(self):
--- a/pombola/core/tests/test_organisation_kinds.py
+++ b/pombola/core/tests/test_organisation_kinds.py
@@ -8,10 +8,22 @@ from django.contrib.contenttypes.models import ContentType
 
 
 class OrganisationKindTest(TestCase):
-    def test_redirect(self):
+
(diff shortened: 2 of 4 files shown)


mhl:
There's the typo in the commit message that you pointed out yourself, and I made a small comment about maybe removing the potentially confusing SingleObjectMixin - otherwise that looks fine.

mhl:
Actually, forget about dropping the SingleObjectMixin comment, what I suggested doesn't work and the variant that does looks uglier to me.

However, I did also notice that the OrganisationKind slug redirects also need to be performed in the case of the position type views, so I added some commits to do them too.

mhl:
Some notes on deploying this - I've run through these checks on the staging site, and it all works as expected.

  URLs to check:

     does
        /organisation/is/political-party/
     work in the first place? (should do)

     does
        /parties/
     work in the first place? (shouldn't do)

  Look at:

     /person/theodore-ahamefule-orji/experience/ 

       will erroneously say: "not a member of any party"

  To fix:

    Deploy latest master, then:

    Change the OrganisationKind's slug from 'political-party' to 'party'
    and create the SlugRedirect objects

        from pombola.core.models import OrganisationKind
        from pombola.slug_helpers.models import SlugRedirect

        ok = OrganisationKind.objects.get(slug='political-party')
        ok.slug = 'party'
        ok.save()

        SlugRedirect.objects.create(new_object=ok, old_object_slug='political-party')

  To check:

      /organisation/is/political-party/
      /organisation/is/party/

      /parties/

      /person/theodore-ahamefule-orji/experience/

PR #1709, opened 2015-07-24
Move Google Analytics JS to static page, rather than lazy-loading.

### LINKED ISSUE #1707: there's no need to load Google Analytics Javascript via Modernizr any more
Rather than having the Google Analytics Javascript snippet in-page as is conventional, it's loaded via Modernizr:

* https://github.com/mysociety/pombola/blob/7ca886ee4e44e2fafdaedfa929c9e81a69aedeb8/pombola/core/static/js/loader.js#L33-L34

@jacksonj04 points out that the Universal Analytics Javascript snippet doesn't block page loading anyway, so there's no reason to do that any more. Switching to including the snippet conventionally would also stop the confusing error messages in Google Analytics saying that it can't find the tracking code (at least, I assume so).

jacksonj04:
GA is not blocking (the original reason for it being lazy), and this allows GA to detect the tracking code and stop complaining.

* Closes #1707

nb This keeps extra print event tracking (in analytics.js) as lazy and compressed, since the `ga` object will be initialised by that point anyway.

### DIFF
--- a/pombola/core/static/js/analytics.js
+++ b/pombola/core/static/js/analytics.js
@@ -1,11 +1,3 @@
-(function(i,s,o,g,r,a,m){i['GoogleAnalyticsObject']=r;i[r]=i[r]||function(){
-(i[r].q=i[r].q||[]).push(arguments)},i[r].l=1*new Date();a=s.createElement(o),
-m=s.getElementsByTagName(o)[0];a.async=1;a.src=g;m.parentNode.insertBefore(a,m)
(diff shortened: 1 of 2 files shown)


mhl:
Looks good :+1: Could I just ask (redundantly, I'm sure) that you double-check that data's still being successfully recorded in GA after deploying this?

jacksonj04:
Deployed and sanity checked, GA is still receiving data.

mhl:
Thanks!

PR #1710, opened 2015-07-24
Remove research survey from Mzalendo

jacksonj04:
Pulls the now complete Quant 1 research survey from display on Mzalendo.

Some elements remain, as they're also used for the ZA survey.


mhl:
I think you'll need to remove:

```
        {% compressed_js 'survey' %}
```

... from `pombola/templates/default_base.html` as well, otherwise that looks fine.

jacksonj04:
The survey code is still used on ZA, but it's removed in #1717, so I'd propose just merging that instead (since it's built on this PR anyway).

PR #1711, opened 2015-07-27
KE: Retain linebreaks in address data using template filters not CSS

### LINKED ISSUE #1700: Display issue on Mzalendo People Pages
screenshot 2015-07-13 15 46 11

lizconlan:
Using 'white-space: pre' causes addresses that are formatted without
linebreaks to be displayed as a continuous line which breaks the
layout of the person pages. This commit uses Django's linebreakbr
filter to force linebreaks in the address text to be rendered as
`<br>` tags instead.

Fixes #1700


mhl:
Looks good :+1:

PR #1715, opened 2015-07-30
ZA: Add search range filter to search

### LINKED ISSUE #1681: pa.org.za - allow filtering of search results by a date range
As per the title - allow user to select from and to dates and only display results that fall within this range

lizconlan:
Adds the date range params and logic to the core search so that it's available for all countries if required but the new UI elements have only been added for South Africa so far.

49e167e also fixes an incidental bug which meant that "Sort by date" was offered to users for the global as well as per-section search but wasn't actually changing the sort order. However this can produce unexpected result ordering where some of the results do not have applicable dates (more detail in commit message).

Fixes #1681

### DIFF
--- a/pombola/core/static/css/jquery.ui.datepicker.css
+++ b/pombola/core/static/css/jquery.ui.datepicker.css
@@ -1,0 +1,68 @@
+/*
+ * jQuery UI Datepicker 1.8.17
+ *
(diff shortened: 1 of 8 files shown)


mhl:
One small suggestion, but :+1: after that - I see what you mean about the search with date ordering being a bit confusing, but I think this is a good improvement nonetheless so we should deploy it and see how it goes.

PR #1721, opened 2015-08-06
NG: Return more detailed info in PUN search

### LINKED ISSUE #1714: PU Search not returning results inline
We've had a message from the team at SYE saying "The PU search function appears to be in place but not effective. Search is allowed, but return isn't in line"

lizconlan:
This changes the PUN search so that it displays whether it matched the PUN at a state, LGA or ward level, and always displays the name of the containing state of the search results.

Adds the name of the current representative for the overlapping districts, where available (or the text 'no current representative' where the representative's name is not held).

Fixes #1714

### DIFF
--- a/pombola/nigeria/templates/search/poll-unit-number.html
+++ b/pombola/nigeria/templates/search/poll-unit-number.html
@@ -26,7 +26,7 @@
 
 {% block search_results %}
   {% if area %}
-    <p>Best match is "{{ area_pun_name }}" with poll unit number '{{ area_pun_code }}'</p>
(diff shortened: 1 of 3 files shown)


lizconlan:
@mhl Scrutinizer links for your entertainment - [~600b1b9 (before)](https://scrutinizer-ci.com/g/mysociety/pombola/inspections/9ea41f7c-8303-44c7-b2b5-388e87dcc626/code-structure/py-class/NGSearchView) and [fbc7180 (after)](https://scrutinizer-ci.com/g/mysociety/pombola/inspections/66828108-fe74-4c47-a49f-ee7388fae3a7/code-structure/py-class/NGSearchView)

duncanparkes:
:+1: Minor tidying up suggestions only :-)

PR #1724, opened 2015-08-13
Code and style redesigned person page

### LINKED ISSUE #1680: Redesign of person page on pa.org.za
Wire framing for now - rough first pass from dave_w below:

pa-mockups 003

wrightmartin:
Closes #1680


mhl:
One small thing is that the biographical text runs right up against the edge of the green box - maybe it could do with some padding / margin on the right?

new-pa-person-pages

mhl:
The test failures are because the "former positions" section has disappeared, e.g. compare:

* http://www.pa.org.za/person/graham-brian-douglas-mc-intosh/

... with:

former-positions-missing

I appreciate that this is another issue with there not being a useful set of example people with different information to test against, but at least in this case there was a test :)

pombola/south_africa/templates/core/person_detail.html
```diff
@@ -7,297 +7,236 @@
...
+      </div>
     </div>
   </div>
-  {% endif %}{% endwith %}
+</div>
+<div class="page-wrapper wrapper">
+  <div class="large-container">
+    <div class="tabs">
+      <ul class="tab-links">
+        {% with past_positions=person.position_set.all.political.currently_inactive %}
```
mhl: It's a bit confusing that this `with` block opening is in here - I'd put it after `large-container`'s reopening just above, and the `endwith` symmetrically just before that `div` is closed.

pombola/south_africa/templates/core/person_detail.html
```diff
@@ -7,297 +7,236 @@
...
-
-  <div id="profile" class="tab-content tab-active">
-
-    <div class="person-summary">
-      {% if person.date_of_death %}
-        <p>Died {{ person.date_of_death }}</p>
-      {% endif %}
-      {{ person.summary }}
-    </div>
+</div> <!-- .large-container -->
```
mhl: These are tiny points, but worth mentioning because it took me a while to figure out: I think it might be worth having a {% comment %} block here just to explain why the enclosing `div`s need to be closed and reopened, since it's a bit surprising to see.  I'd also make sure that the indentation is such that the block looks like:

```
{% block subcontent %}
    </div> <!-- .large-container -->
  </div> <!-- page-wrapper -->
  ...
  <div class="page-wrapper wrapper">
    <div class="large-container">
    ...
{% endblock %}
```

wrightmartin:
> The test failures are because the "former positions" section has disappeared, e.g. compare:

Former positions are now under the 'positions held' tab, I guess this means the test needs updating?

mhl:
Hi @wrightmartin - I've fixed the tests now, but I'm afraid spotted another couple of things that might be issues:

* The date of death is shown in the profile box for people who are deceased, but the date of birth isn't shown (e.g. there should be a date of birth for both `/person/elizabeth-thabethe/` and `/person/hlakudi-frans-nkoana/`
* If someone has no "Appearances" or "Interests Register" data it looks a bit confusing because the "Positions" tab isn't made active, e.g. for `/person/maria-tshabalala/`, the page initially looks like:

on-page-load

... which looks like they have no positions, but if you click on the tab, you get:

after-clicking-tab

I can fix this one if you don't have time or it's not clear how to.

PR #1725, opened 2015-08-14
ZA: Add Facebook comment support

### LINKED ISSUE #1731: Complete Facebook comments integration on pa.org.za
Add to constituency office pages
Have archived comments visible on the same page (read only)

lizconlan:
Adds `ZA_FACEBOOK_APP_ID` as a config setting then uses that in
preference to any Disqus config values to pull in the Facebook
comments plugin instead.

The Facebook equivalent of the comment thread identifier has to
be a fully formed URL and is currently set using
`request.build_absolute_uri`. However, the url doesn't have to match
that of the page it is on, which could be useful if we wanted to
include the same discussion on multiple pages.

A word of caution - changing which site the comments are associated
with in the Facebook app (and the domains have to match or it won't
work) seems to lead to the original comments being lost.
 
The disqus.json file contains a recent copy of all the Disqus posts, retrieved via the Disqus API (by calling `https://disqus.com/api/3.0/forums/listPosts.json?forum=peoplesassembly&order=desc&related=thread&limit=100` using a login which is not authorised as an administrator) rather than calling the API from every blog and organisation page, especially as the plan is to close all the comment threads once the Facebook system is live so the information will not change after we've done that. __The file will need to be updated once the comment threads have been closed to make sure we have picked up any new comments__

Comments on pages which have pre-existing Disqus comments will look like this (but without the 'Join the discussion...' input box once the threads have been closed to new comments):

screen shot 2015-09-02 at 11 12 47

Note - if a page's comments aren't appearing in Facebook's moderation tool, it might be that it's not (yet) picking up the appId properly. Use the [Object Debugger tool](https://developers.facebook.com/tools/debug/og/object/) to force the issue by "rescraping" it (h/t for that, as ever, to [a StackOverflow response](http://stackoverflow.com/a/12648301))

Fixes #1731


mhl:
Quick question - do you know how we close all the Disqus comment threads for the site?

mhl:
Quick question - do you know how we would close all the Disqus comment threads? (e.g. is there an easy way in the Disqus admin to mark them all as read-only or something?)

mhl:
Goodness, it shows what kind of day I'm having when I can't remember that I submitted the same comment *two minutes previously*

lizconlan:
Irritatingly, I can't see a way to do it through the admin console (on my account, looking at a test site I created - I don't have admin rights for PMG).

I *think* you can do it at the "forum" level - i.e. close everything in the container for the site's threads -via the API using this:

https://disqus.com/api/docs/threads/close/

(If not then it can be done by an API query for all the open threads, collect the IDs and close them one by one)

Or as a slightly less certain method, one of the options in the admin settings is to autoclose the threads after *n* days (set to 0 by default to indicate "do not do this, ever") so it could be set to 1 day and let nature take its course as we won't be creating any new ones.

mhl:
I think my only comment about this is that I think it would be worth making the option `FACEBOOK_APP_ID` and having the same logic about displaying Facebook and / or Disqus comments available in all Pombola countries, on the basis that:

* Other countries might well want the same change if this is successful
* In general, it's best to try to minimise the amount of country-specific code

What do you think?

lizconlan:
@mhl I agree, slowly moving things around

mhl:
Great, this looks fine - I think we should leave deploying this until Monday; I'll merge and deploy it then. Thanks!

lizconlan:
possible enhancement - if, when the thread is closed, the list of other Disqus posts is still pulled in, we could hide it with some CSS magic, something like:

    #disqus_thread .post-list {display: none}
    #disqus_thread #thread-subscribe-button {display: none}

(the second one also hides the subscribe button, unsurprisingly, although hopefully that will go automatically as it doesn't make any sense to have it for a closed thread)

wrightmartin:
I've got a small concern about performance here - I don't know how big FB comments are, but Disqus adds at least 200kb to the payload. 

Can we possibly load disqus on demand? Maybe hide it under a button to show archived comments if the visitor wants to see them? 

The use-case for viewing archived comments must be small?

mhl:
This has been merged (as 74ae683 and ancestors), so I'm closing the PR

PR #1726, opened 2015-08-14
Include place data for positions in the Popolo JSON output

### LINKED ISSUE #1688: include associated Place data in Popolo export
`core_export_to_popolo_json` should be modified to include Place data where it's attached to memberships - for example, places attached to Mzalendo member of parliament positions are specific to a parliamentary session, and this data may well be useful to EveryPolitician.

mhl:
Some positions (e.g. those representing an MP for a particular
constituency) have a place attribute. This wasn't being exported
in the nightly JSON export, however. This commit introduces a change
so that the place data is exported as the 'area' attribute of a
Popolo membership.  In addition, if the place is specific to a
particular parliamentary session (as happens with constituencies in
Mzalendo) the parliamentary session data is included under the
'session' attribute of the area.

Fixes #1688

### DIFF
--- a/pombola/core/popolo.py
+++ b/pombola/core/popolo.py
@@ -6,8 +6,11 @@ from collections import defaultdict
 from urlparse import urljoin
 
 from django.core.management.base import BaseCommand, CommandError
+from django.core.urlresolvers import reverse
 from django.db import transaction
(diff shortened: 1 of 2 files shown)


lizconlan:
:+1: looks good

PR #1738, opened 2015-08-21
Better Hansard speech reattribution functionality

### LINKED ISSUE #1712: Mzalendo 2013 Data corrections
@duncanparkes @paullenz @mhl 

I. Emma Getrude a nominated Senator has two existing profiles, Kindly merge the two.
a)http://info.mzalendo.com/person/emma-mbura-gertrude/
b)http://info.mzalendo.com/person/emma-mbura-getrude/

2. Aaron Rotich Tuikong's speech entries should be moved to Hon Mary Emaase's profile. Rotich is not a member of parliament. 

3. The speech entries on Josephat Nanok's profile from March 2013 should be moved to Daniel Epuyo Nanok's profile.

4. The speech entries on Eric Opon Nyamunga's profile from March 2013 should be moved to Rose Auma Ogendo's profile.

5. The speech counts under James Mungai Njoroge should be moved to Nakuru's Senator profile James Kiarie Mungai. James Mungai Njoroge erroneously appears as a Member of Parliament. 

6. The speech counts under Rachael Nduku Ndambuki should be put under Rachael Kaki Nyamai. Rachael Nduku Ndambuki is not a member of parliament.

lizconlan:
Fixes a bug in the `core_merge_people` management command which was preventing it from picking up hansard speeches and interests register information. Also provides more data to the user and offers them the choice of whether or not to continue once both speakers have been identified (with a specific warning if both speakers turn out to be the same person - i.e. if one is an alias of the other).

Adds a new management command, `hansard_reattribute_entries` which takes 2 person identifers and optional to and from dates in order to assign one person's speeches to another. Does not attempt to delete the donor speaker from the database even if this leaves them with no contributions.

Both commands will now accept either a numeric person id (as before) or a slug.

Partially fixes #1712 (won't be properly fixed until the updated commands have been run on the live site and the csv files have been regenerated correctly)

### DIFF
--- a/pombola/core/management/commands/core_merge_people.py
+++ b/pombola/core/management/commands/core_merge_people.py
@@ -148,7 +148,7 @@ class Command(PersonSpeakerMappingsMixin, BaseCommand):
         # Then those in hansard, if that application is installed:
         #    hansard_models.Alias
         #    hansard_models.Entry
-        if 'hansard' in settings.INSTALLED_APPS:
+        if 'pombola.hansard' in settings.INSTALLED_APPS:
             import pombola.hansard.models as hansard_models
             hansard_models.Alias.objects.filter(person=to_delete).update(person=to_keep)
             hansard_models.Entry.objects.filter(speaker=to_delete).update(speaker=to_keep)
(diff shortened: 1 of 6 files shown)


mhl:
Sorry, lots of comments for a pull request that's essentially fine! (and :100: for the thorough and easy-to-read tests). I hope the comments are constructive, anyway.

The other alternative to using `raw_input` for confirmation of what the command is doing would be to add `--commit` option (which makes it easier to use the command in a script, but people might still have that "do I really want to run this command?" feeling if they don't know about it). Alternatively you could add a `--noinput` option to skip the `raw_input` prompt, as some of core Django's admin commands have - I have a slight preference for the latter in this case.

PR #1741, opened 2015-08-24
Add a featured_image_file property to InfoPage for thumbnails of posts

mhl:
The new design for the People's Assembly home page shows thumbnails of
recent infographics.  This should be simple to add, but, at the moment,
the infographic images in an blog post are just copy-and-pasted links
inserted into the HTML or Markdown that point to files that have been
uploaded with the file_archive application.

In order to display these efficiently, I'm adding a featured_image_file
field to InfoPage, and the data migration introduced in this commit
takes the first image from each post (if it's Pombola-hosted) and sets
it as the featured_image_file.


duncanparkes:
:+1:

duncanparkes:
Good spot with the URL possibly containing a domain.

:+1:

PR #1742, opened 2015-08-26
1719 name matching algorithm settings

### LINKED ISSUE #1719: add feature flag for Ghana parser changes
> As a developer of Pombola
I want to be able to merge in the current Odekro site without breaking the current Hansard parser
So that we can have the code for Odekro in Pombola's master branch

The `merge-current-rebased` branch, which brings in the current version of Odekro into Pombola, is nearly ready, but currently does change the Hansard parser to use a different name matching algorithm. This would most likely break the parser for Kenya, so we should add a feature flag (set from the countries' base settings files) to select one algorithm or the other. The differences in the parser can be seen from:

    $ git diff origin/master origin/merge-current-rebased -- pombola/hansard/models/entry.py

duncanparkes:
Add a setting to allow the Hansard name matching algorithm to be changed per country - this means Kenya can carry on doing it the old way, while we use the new improved algorithm for Ghana.

Fixes #1719

### DIFF
--- a/pombola/hansard/constants.py
+++ b/pombola/hansard/constants.py
@@ -1,0 +1,4 @@
+# Symbols for different name matching algorithms
+
+NAME_SUBSTRING_MATCH = 1  # Look for the whole name without the title.
+NAME_SET_INTERSECTION_MATCH = 2  # Intersect sets of words, including the title.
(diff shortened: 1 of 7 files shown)


mhl:
As we discussed in IRC, there are actually some test failures if `KENYA_PARSER_PDF_TO_HTML_HOST` is set (otherwise the existing test for `assign_speakers` isn't run) - these would be worth investigating, although they may have been present before this pull request, since I typically haven't had that setting set.

Ideally, I'd like to get rid of that setting completely; replacing it with a path to the `pdftohtml` binary to use would be much more helpful.

mhl:
:+1: I think this should be merged into `merge-current-rebased` now.

Issue #1670, opened 2015-06-15
Redesign of the pa.org.za homepage


paullenz:

paullenz:
Wireframe in the first instance - very rough first pass done by dave_w when in ZA show below.

+ some notes:  


1. The curated issue pages we have to put on hold as a possible phase 2 as we do not have the HR currently
Could we replace the 6 buttons with the topics below:
 
Middle 6 buttons:  MP comments* / MP Corner / Minister's Replies to MPs / MP Attendance / MP Assets / Activist Campaigns
* Could one have a search button that takes you to a dedicated page showing just the MP's Committee and Plenary comments? These appearances would still remain as well on the MP's profile page but it would be a way to surface another part of the website, Appearances, that is pure technical wizardry.

![pa-mockups 001](https://cloud.githubusercontent.com/assets/909811/8282617/e44a66d6-18e9-11e5-816b-cce5807c3c79.png)

zarino:
New homepage layout, retaining the current PA style:

![homepage - wide](https://cloud.githubusercontent.com/assets/739624/8356477/7a372b9c-1b4d-11e5-9e59-ae6b464a4101.png)

(This is what it might look like on a narrow screen:)

![homepage - narrow](https://cloud.githubusercontent.com/assets/739624/8356480/83127686-1b4d-11e5-88bb-63a66f026efd.png)

Here it is in full green, although I'd suggest against it for two reasons:

* It's harder to differentiate between normal text and links, since, aside from white, not very many colours work well on dark green.
* Maybe it's just me, but it feels quite oppressive. Something to test with local users maybe?

![homepage - wide dark](https://cloud.githubusercontent.com/assets/739624/8356582/1b0e1634-1b4e-11e5-9274-915ef4c51a5f.png)

If they *really* want the green, but agree a full green background is too much, we could always tweak the existing style to make green the primary accent colour:

![homepage - wide light](https://cloud.githubusercontent.com/assets/739624/8356621/66ab4d64-1b4e-11e5-89ae-be628cbc964e.png)

That's only a really quick mockup – we'd probably work a little harder on getting the colours right, but it gives you an idea.

davewhiteland:
Awaiting feedback (@paullenz emailed wireframes): PMG said they like them :-) and that Zaheedah is handling this but has been away (2-Jul-15), so expect followup soon...

wrightmartin:
[...]

PR #1753, opened 2015-09-02
ZA: Improve homepage spacing on short but wide screens

zarino:
https://github.com/mysociety/pombola/issues/1670#issuecomment-136663910 raised the issue of the infographics on the homepage being too low down, and being cut off on short desktop screens.

To some extent this can't be helped for ultra-short screens (ie: 600px high or less) – the People's Assembly header and nav bar alone take up 150px.

But I've attempted to improve the situation for screens in the middling category – somewhere between 600px and 800px high. For a drastic comparison, here's before (left) and after (right), at 650px high:

700px high:

And 750px high:

Note how more of each infographics is visible on page load, so people know they can scroll down.

### DIFF
--- a/pombola/south_africa/static/sass/_south-africa_overrides.scss
+++ b/pombola/south_africa/static/sass/_south-africa_overrides.scss
@@ -610,14 +610,9 @@ body.home {
 .home__section {
     padding: 1.25em 0.75em; // roughly 20px and 10px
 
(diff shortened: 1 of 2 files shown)


Gemmamysoc:
Have passed to Megan for feedback :)

mhl:
:+1:

Issue #1737, opened 2015-08-21
[NG] Governor's page returning a 404 error


JenMysoc:
http://www.shineyoureye.org/position/governor/ is returning a 404 error when it should list the Executive Governors of States

mhl:
*after quite a bit of investigation, and thanks to the waybackmachine and the Django admin log...*

I believe what's happened here is that on the 22nd of April 2015, the `eienigeria` user edited the `governor` position title in the admin interface to have the slug `executive-governor` instead. This meant that the link http://www.shineyoureye.org/position/governor/ no longer finds that position title - someone needs to update it to http://www.shineyoureye.org/position/executive-governor/

It might be worth passing on the message that changing the slug of a position title in the admin interface may well break links on the site. (And / or we should make certain fields non-editable if changing them would reak links on the live site.)

duncanparkes:
There's a general problem here in Pombola that in lots of places this kind of minor thing about the data is important and baked into the code. The long term solution to this is to use a different property of a position to determine whether it appears in a special place in the UI from that used to determine its URL, but for now, yes, we just have to change it back, I think.

duncanparkes:
In order to preserve both urls for the governor position title, I've allowed slug redirects for position titles in #1754.

duncanparkes:
If #1754 is accepted, then we need to create a SlugRedirect object to point 'governor' at 'executive-governor', or change the slug back and have a redirect the other way around (which would probably be my preference).

duncanparkes:
I've made added the SlugRedirect, and checked it's working, so this can now be closed.

PR #1754, opened 2015-09-03
Allow slug redirects for PositionTitles.

duncanparkes:
Allow position title urls to use slug redirects. This means that if someone edits a slug of something for which the slug is baked into the templates, we can keep all the urls that were ever in use pointing at the right place.

This is the code part of the fix for #1737.

Connect to #1737


duncanparkes:
Connect to #1737

PR #1758, opened 2015-09-04
ZA: More indices in pagination links

### LINKED ISSUE #1652: ZA: show more page indices in pagination links
(Raised by PMG: this at least applies to committee appearances and blog posts.)

pagination-1

pagination-2

In both these cases, they'd like more indices (at least 4 at the start and end, I guess) in the pagination links.

lizconlan:
Shuffles the app loading order so that 'pagination' is loaded later and we get the option to override the default `pagination.html` on a per country basis (e.g. by adding pombola/south_africa/templates/pagination/pagination.html)

Changes South Africa's `PAGINATION_DEFAULT_WINDOW` from 2 to 3

Changes the pagination style of the blog posts from

screen shot 2015-09-08 at 17 29 44

to

screen shot 2015-09-08 at 17 19 03

Unfortunately, changing the text from PREVIOUS and NEXT back to NEWER POSTS and
OLDER POSTS is non trivial as it looks as though [the way django-pagination includes the pagination.html template](https://github.com/ericflo/django-pagination/blob/master/pagination/templatetags/pagination_tags.py#L245-L246) only allows a single override per running application so it's all or nothing. Also it fails to pass the full request or context back into the template which hampers any attempts either to pass in something like `context['prev_text']` or to work out what the current url path is to override the text (e.g. if it contains '/blog/').

The only ways to fix this I can think is patching django-pagination to pass the context object to the template or copying [the logic that sets up the pages list](https://github.com/ericflo/django-pagination/blob/master/pagination/templatetags/pagination_tags.py#L151-L221) into a Mixin and bypassing django-pagination altogether, at least for the blog list pages.  

Fixes #1652


dracos:
You can (if I've understood it right) drop pombola/templates/pagination/pagination.html as it'll never be found – not sure it was ever found before, in fact, as the loaders are set to Apps before Filesystem (since https://github.com/mysociety/pombola/commit/09640b660c38bcaf286b827c37dffb9f815e38c4), so it would always find the template in the pagination app and never that one.

lizconlan:
@dracos - that sounds like what I was seeing when trying to get the text to change, will add it to the list

mhl:
Looks good - the other page that it strikes me could do with better pagination is search results. Would you have time to look at switching that to use django-pagination as well, possibly?

mhl:
:+1: Great, thank-you

Issue #1764, opened 2015-09-14
front-end performance sprint


mhl:
The idea of this is to spend a few days (Monday 21st of September to Wednesday 23rd of September) working with @wrightmartin and / or @zarino to focus on getting down page-load times for Pombola sites, and look at ways we could improve the experience for people using the site by reducing asset sizes, changing load order, using image sprites, etc.

It would also be a good chance to look at questions like "is it still worth using Modernizr?", related to https://github.com/mysociety/pombola/issues/1686

wrightmartin:
- Shall we give ourselves a target? Say ~10% of requests, ~20% filesize?
- We shouldn't worry about any of the more difficult optimisations (critical path rendering etc) until we've taken care of the low hanging fruit
- We should make this process as repeatable as is sensible, so we could do it at regular intervals (every year?) and apply it to all of our sites.
- We should draw up rough guidelines for new projects, so that we always start from a performant place

wrightmartin:
### Pombola specific
- [ ] Ensure we're only loading CSS and JS that is used for the current domain (Eg. no mzalendo assets loading on People's Assembly) 
- [ ] Ensure we're only loading assets of 'one-off' pages on the page they are needed for (Eg. Mzalendo women's page)
- [ ] Look at jQuery dependencies - try and use native JS if possible
- [ ] Consider replacing or removing social media plugins

### General performance enhancements
- [ ] Combine and gzip all JS files
- [ ] Image compression
- [ ] Ensure images are rendered at the size they are required (could use ```srcset``` for this)
- [ ] Minimise requests
- [ ] Use sprites for CSS images

wrightmartin:
### Mzalendo

Homepage
![screen shot 2015-09-21 at 09 04 49](https://cloud.githubusercontent.com/assets/2292925/9987455/e8580b0a-603f-11e5-928e-d8816b39fe0d.png)

Person page
![screen shot 2015-09-21 at 09 05 06](https://cloud.githubusercontent.com/assets/2292925/9987457/ed6008dc-603f-11e5-9447-1719b3705608.png)

wrightmartin:
People's Assembly

Homepage
![screen shot 2015-09-21 at 09 06 06](https://cloud.githubusercontent.com/assets/2292925/9987482/216e71fe-6040-11e5-9705-66278ae90160.png)


Person page
[...]

PR #1769, opened 2015-09-21
Recompress all images

wrightmartin:
I've run every image asset I could find through ImageOptim, twice

> Saved 868.2KB out of 3.1MB. 27.6% overall (up to 91.7% per file)


wrightmartin:
Part of #1764 1764

mhl:
:+1:

PR #1775, opened 2015-09-21
KE: Use srcset on homepage parliament image

zarino:
Reduces download overhead, for that single image, on mobiles and desktops, by about 50%. We use srcset to serve a slightly larger image for devices between 400px and 640px width.

## Before (104KB file regardless of screen size)

screen shot 2015-09-21 at 12 41 52

## After – mobile (49KB file)

screen shot 2015-09-21 at 12 38 59

## After – mobile-to-phablet transition* (49KB + 64KB files)

\* This is a worst-case scenario: someone loads the page on phablet, and then rotates it, crossing the 400px boundary. A second version of the image downloads in a size suitable for the new screen resolution.

screen shot 2015-09-21 at 12 39 18

## After – phablet (64KB file)

screen shot 2015-09-21 at 12 40 55

## After – tablet and up (49KB file)

screen shot 2015-09-21 at 12 39 58


dracos:
What are the thoughts on image quality? I don't have the full size to start with, but dropping the 600px down to 400px and reducing quality to 70 and 50 gives 41K and 30K respectively, is there a noticeable difference with the 49K one? Here's all three side by side:
parliament-400px-50
parliament-400px-70

zarino:
@dracos – just pushed a new version of the images: 30KB for the 400px one and 55KB for the 600px one. There are visible artifacts, but only really if you're looking for them.

parliament-400px

parliament-600px

zarino:
Also worth mentioning the images are progressive JPEGs which will hopefully make the page feel faster loading too.

dracos:
Woo :)

PR #1776, opened 2015-09-21
ZA: Optimize header logo

zarino:
Removes the tagline text from the header logo, and brings high-dpi image sizing in line with how we do it elsewhere.


mhl:
:+1:

PR #1778, opened 2015-09-21
ZA: Crop and compress homepage infographic previews

zarino:
The four infographic previews on the homepage are cropped so that only the top 250px are included, and also given 80% JPEG compression (down from the default 95%).

On my VM, this represents a roughly 90% saving in file size (from 120KB each image, down to 14KB).


mhl:
:+1:

PR #1779, opened 2015-09-21
NG: Reduce size of Nigeria homepage spinner images

zarino:
Same code as #1778. The spinner images are generally displayed at 250px or less so there's no point in making the client download the current 478x478px images.

As it stands, those images are the largest files on the homepage, weighing in at 40–60KB each. This should reduce those sizes by roughly 50%.


mhl:
:+1:

PR #1781, opened 2015-09-21
Add DNS prefetch tags for Disqus and Twitter Javascript

mhl:


pombola/core/templates/core/organisation_base.html
```diff
@@ -3,6 +3,13 @@
 
 {% block title %}{{ object.name }}{% endblock %}
 
+{% block extra_head_meta %}
+  {# Prefetch DNS for when the Disqus Javascript is loaded #}
+  {% if settings.DISQUS_SHORTNAME %}
+    <link href="//{{ settings.DISQUS_SHORTNAME }}.disqus.com" rel="dns-prefetch" />
+    <link href="http://{{ settings.DISQUS_SHORTNAME }}.disqus.com/" rel="prefetch" /> <!– IE9+ –>
```
dracos: Did you have some auto-correct of -- happening ;) I think you need it for IE9 only, according to http://blogs.msdn.com/b/ieinternals/archive/2012/03/01/ie10-beta-consumer-preview-minor-changes-changelist.aspx so you could wrap the line in `<!--[if IE 9]>…<![endif]-->` so only IE9 sees it and others don't fully prefetch that page.

mhl: Oops, thank-you - I'll force-push a new version just now

PR #1784, opened 2015-09-22
Add SVG of People's Assembly logo

wrightmartin:
This also adds in an SVG support check to core (which we may or may not need)


zarino:
Out of interest, what's the reasoning behind setting both `display` and `visibility` on the various items in [lines 383–416 of _south-africa_overrides.scss](https://github.com/mysociety/pombola/blob/3a67082d0485c2ac0554a40a6b84df13b28456ed/pombola/south_africa/static/sass/_south-africa_overrides.scss#L383-L416)? Wouldn't `display` do just fine on its own?

zarino:
I'm going to assume that the inline SVG is going to be compressed by gzip (running gzip locally on the file reduced it from 11KB to 3.5KB) but I'm going to mention it anyway just in case – are we ok with adding potentially 3KB to the markup of every page on the site, especially when it can't be cached between pages?

wrightmartin:
> Out of interest, what's the reasoning behind setting both display and visibility on the various items in lines 383–416 of _south-africa_overrides.scss? Wouldn't display do just fine on its own?

So I understood it that `visibility: hidden` had more value for screen readers, ie they would ignore the contents, whereas that's not always the case with `display: none`. However I can't find anything on the Googles to confirm this, so I'll remove it.

wrightmartin:
>I'm going to assume that the inline SVG is going to be compressed by gzip (running gzip locally on the SVG markup reduced it from 11KB to 3.5KB) but I'm going to mention it anyway just in case – are we ok with adding potentially 3KB to the markup of every page on the site, especially when it can't be cached between pages?

Yes, we lose a request but we lose caching, and we'll shave a bit off the filesize from the initial load. Which is preferable is up for debate....

wrightmartin:
But there are some good stats here

http://stackoverflow.com/questions/9855545/http-requests-vs-file-size

zarino:
> Yes, we lose a request but we lose caching, and we'll shave a bit off the filesize from the initial load. Which is preferable is up for debate....

*TL;DR: Well this has been a fun little diversion. I wondered whether maybe we could solve the debate with data from Google Analytics. Turns out maybe we can, but only if know how many people we're willing to inconvenience for the benefit of the majority…*

The inline SVG is only really a problem for people who visit multiple pages (where, if the SVG hadn't been inlined, it would most likely still be in the browser cache for the second and subsequent pageviews). 

Looking at Peoples Assembly in particular, with data from 1st Jan 2015 to yesterday, 77% of sessions included only a single page view, 10% viewed two pages, 5% viewed three pages, 2% viewed four pages, and so on.

I've just done a quick test locally on my Mac, and the gzipped SVG on its own is 3.5KB, the homepage HTML (as it is on the current live site) is 5.4K gzipped, and the homepage with the SVG inserted inline is 8.7K. So HTML+SVG as a single file turns out to be 200 bytes smaller than when delivered as two separate files.

If we assume users view a succession of pages of a similar weight to the homepage, here's how it'd work out:

| number of pages viewed | total with inline SVG | total with separate SVG |
|---|---|---|
| 1 page | 8.7KB | 8.9KB |
| 2 pages | 17.4KB | 14.3KB |
| 3 pages | 26.1KB | 19.7KB |
| 4 pages | 34.8KB | 25.1KB |

So assuming we inline the SVG – out of a typical sample of 100 visitors, 77 will actually get what they came for having downloaded 0.2KB *less* than if the files were separate. So that's a saving of 15.4KB across those 77 users. 10 users will download 3.1KB *more* than if the files were separate, meaning a total of 31KB overhead for them as a group. 5 users would download 6.4KB *more*, meaning a total overhead of 32KB for them. And for simplicity, let's assume the final 4 (rather than 2 as analytics suggest) download four pages each; that'd be an overhead of 9.7KB each, or 38.8KB as a whole.

In total, considering all 100 users, some of whom saved bytes, and some of whom downloaded extra bytes, inlining the SVG costs our 100 users 86KB in total.

So if we're considering file size only, inlining the SVG is a worse experience for the group as a whole but *really* it only costs the 23% who view more than one page. How much do we care about the 23%?

What I haven't considered is the benefit of reducing HTTP requests. Just because I'm not sure how to measure the impact of requesting the webpages with and without an extra HTTP request. Maybe someone more geeky than me could have a stab at calculating it, I'd love to see the findings.

PR #1790, opened 2015-09-23
There's no need to use JS to unlink the current sub-page menu link

mhl:
Edmund (@evdb) left this comment in js/both-functions.js, introduced in
f0ec7ef5ff372398:

// In the ${object}_base.html templates there are menus linking to the subpages.
// It proved to be difficult to elegantly get these to render differently on
// each page so that the current page link was not a link, or muted.  Mainly this
// was because to avoid repeating complicated if locks in the templates I wanted
// to use an include to generate the li element. But that meant passing in the
// name of the page to load for {% url foo ...%} and that can't take foo as a
// variable until Django 1.5. Sigh.
//
// Hence writing a teeny bit of js to do it instead.

Now we're on a version of Django >= 1.5 we can do this server-side
instead.  This commit makes those changes.

This allows us to remove js/both-functions.js completely.


pombola/core/views.py
```diff
@@ -128,6 +144,12 @@ def get_context_data(self, **kwargs):
         context = super(BasePlaceDetailView, self).get_context_data(**kwargs)
         context['place_type_count'] = models.Place.objects.filter(kind=self.object.kind).count()
         context['related_people'] = self.object.related_people()
+        if settings.ENABLED_FEATURES['projects']:
+            # The number of projects associated with the place is used
+            # in the link text in the object_menu_links:
+            context['projects_sub_link_text'] = 'CDF Projects ({0})'.format(
```
dracos: I don't know if translation is on the horizon, but in general better to use `%(number)s` interpolation because that's clearer to the translator.

pombola/core/views.py
```diff
@@ -96,12 +96,28 @@ def redirect_to(self, correct_object):
...
 
-class PersonDetail(SlugRedirectMixin, SkipHidden, DetailView):
+class BasePersonDetailView(SkipHidden, DetailView):
     model = models.Person
 
+    def get_context_data(self, **kwargs):
+        context = super(BasePersonDetailView, self).get_context_data(**kwargs)
+        if settings.ENABLED_FEATURES['hansard']:
+            hansard_count = self.object.hansard_entries.count()
+            if hansard_count:
```
dracos: I think this if statement is the wrong way round.

mhl: Oops, yes, thanks :(

pombola/core/views.py
dracos: And in Django 1.9 the `blocktrans asvar foo` I added arrives, which means this could be back in the template ;)

dracos:
Looks fine; the deleted code includes a comment that it "keep it a link so that the padding etc does not change." whilst the new code does remove the link but no CSS changes; I assume the display of the links is still okay :)

pombola/core/views.py
mhl: That's reminded me - I've just put i18n of Pombola on the agenda for discussion next week. I'll change it to `'...{number}...'.format(number=...)`

Hooray for `blocktrans asvar` then! It does seem silly that this ends up in the view.

dracos: I guess as it's only a number at the end, you could translate the "CDF Projects" and concatenate on the number, all in the template. Probably not worth it at the mo!

mhl:
Yeah, I couldn't see any problem with the display, e.g.:

tabs-delinked

... so thought I'd leave it until someone with more aesthetic discrimination complains :)

PR #1793, opened 2015-09-23
Use srcset for people photos

wrightmartin:
This affects all instances of Pombola that don't override this part of `object_base.html` (all except Mzalendo, as far as I can tell)


mhl:
:+1:

PR #1794, opened 2015-09-24
ZA: fix the astonishingly slow display of blog pages

mhl:
In combination, these commits take the time for the SQL on blog pages
down to 0.1 second from about 30 seconds.

### DIFF
--- a/pombola/south_africa/templates/info/_blog_sidebar.html
+++ b/pombola/south_africa/templates/info/_blog_sidebar.html
@@ -14,12 +14,14 @@
       {% endfor %}
     </ul>
 
-    {% if popular_posts.exists %}
+    {% if some_popular_posts %}
       <h3>Popular posts</h3>
 
       <ul>
         {% for post in popular_posts|slice:":5" %}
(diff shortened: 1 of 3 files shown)


pombola/south_africa/templates/info/_blog_sidebar.html
```diff
@@ -14,7 +14,7 @@
       {% endfor %}
     </ul>
 
-    {% if popular_posts.exists %}
+    {% if some_popular_posts %}
```
dracos: DO you want to have an `{% empty %}` section here too, now that it's theoretically possible to get here but not display anything? (If all the blog posts are old, for example.)

dracos: Or I guess just the date bit could be in the first query if that doesn't slow it down.

dracos:
Wow, that was slow. Good work :)

pombola/south_africa/templates/info/_blog_sidebar.html
mhl: I've just added an empty block; I think it'll very rarely be seen so probably not a big deal anyway

PR #1802, opened 2015-10-07
Update sort order for position views and fix MP profiles link

### LINKED ISSUE #1801: ZA: MP profiles link is unhelpful
At the moment this just links to a list of all people in the site, including old election candidates who were never elected, etc. It should just link to members of parliament.

mhl:
Fixes #1801

### DIFF
--- a/pombola/core/views.py
+++ b/pombola/core/views.py
@@ -325,7 +325,7 @@ def position(request, pt_slug, ok_slug=None, o_slug=None):
     # Order by place name unless ordering by person name is requested:
     order = 'place'
     if request.GET.get('order') == 'name':
-        order = 'person__legal_name'
+        order = 'person__sort_name'
     positions = positions.order_by(order)
 
     positions = positions.select_related('person',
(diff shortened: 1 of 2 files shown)


pombola/core/views.py
```diff
@@ -325,7 +325,7 @@ def position(request, pt_slug, ok_slug=None, o_slug=None):
     # Order by place name unless ordering by person name is requested:
     order = 'place'
     if request.GET.get('order') == 'name':
-        order = 'person__legal_name'
+        order = 'person__sort_name'
```
duncanparkes: `sort_name` is obviously the right thing, but it can be blank? Can we rely on the data here?

duncanparkes:
:+1:

pombola/core/views.py
mhl: We make sure that when a model is saved, the sort name is set to a (somewhat) sensible default here: https://github.com/mysociety/pombola/blob/master/pombola/core/models.py#L506-L512 and when the field was introduced the migration set sort_name for everyone: https://github.com/mysociety/pombola/blob/7ca886ee4e44e2fafdaedfa929c9e81a69aedeb8/pombola/core/migrations/0055_auto__add_field_person_biography__add_field_person_national_identity__.py#L98-L103 so it should always have a value

Issue #1807, opened 2015-10-12
ZA: PMG's requested changes to the person page on PA


mhl:
After our call with PA last week we figured out some changes that would improve the person pages on People's Assembly; this ticket is for the work to implement these changes (largely design time).

@zarino's notes from the call are here: https://docs.google.com/document/d/14fSsB_ZpyXurzK72XdV2MqORylT7p6mahOjgYtlWLas/edit (only shared mySociety internal)

And a mockup from @wrightmartin of what they might look like: ![mockup of new people pages](http://i.imgur.com/9Lc9Cvl.png)

zarino:
Pull request here: https://github.com/mysociety/pombola/pull/1808

PR #1808, opened 2015-10-19
Peoples Assembly person page tweaks

zarino:
Fixes for #1807.

Sorry for the seeming grab-bag of changes. I’ll try to summarise them here, but if you think some squashing/splitting of commits would make it clearer, go ahead and do it.

#### Changes specific to Peoples Assembly person pages:

1. Move the person details back into the sidebar (like they are in core) and remove styling for the horizontal green section that used to hold those details.
2. Show region (rather than a list of memberships) under the person's name.
2. Tweak the display of the "Appearances" section, as per PMG's feedback:
    * Order: committee appearances, then questions, then plenary
    * Number: 5 committee appearances, 3 questions, 2 plenary attendances
    * If questions list is empty show "No questions found" rather than "No speeches found"

#### Changes for all cobrands:

1. On the desktop view of the person page, replace the absolutely-positioned layout with floats and CSS `calc()`. This should make it easier to maintain and tweak in future, as well as getting rid of the `#hfWrap` javascript hack that set the content height on page load. Desktop browsers that don’t support `calc()` (0.5% of Mzalendo visitors, 5% of Peoples Assembly) get a percentage-based width, which looks *acceptable* if a little weirdly spaced.
1. Reduce CSS specificity in the core person page styles (mostly by removing the `#page` selector that almost everything was nested inside) and then update the Peoples Assembly and Mzalendo styles to benefit from this change (removing lots of `#page` mostly – huzzah!).
2. Also, after discussion with @tmtmtmtm @mhl in IRC, add a few classes to the information in the person page sidebars, to make it easier for scrapers to work out what's a party link and what's a coalition link.

I’ve eyeballed the changes in all five cobrands (kenya, south-africa, ghana, nigeria, zimbabwe) but since I don't have realistic data locally, it'd be good to see this deployed to staging servers for each cobrand, to check there aren't any problems.

### DIFF
--- a/pombola/core/static/sass/_page.scss
+++ b/pombola/core/static/sass/_page.scss
@@ -67,14 +67,8 @@
   width: 100px;
 }
 
-.profile-pic {
-  img {
-    width: 100px;
-    height: 100px;
-  }
-}
-
-#map_canvas {
+#map_canvas,
+.profile-pic img {
   width: 100px;
   height: 100px;
 }
@@ -224,11 +218,13 @@
   }
 
   .title-space {
-    min-height: 0;
-    float: left;
-    clear: left;
-    margin: 0 0 20px 240px;
+    min-height: 0; // everything's floated now, so no need to "clear" the absolutely positioned .profile-pic
+    margin: 0;
     position: relative;
+
+    float: right;
+    width: 60%; // fallback for browsers that don't support calc()
+    width: calc(100% - 240px); // parent width minus 240px for sidebar
 
     .object-kind {
       font-style: normal;
@@ -239,27 +235,21 @@
     }
   }
 
-  .profile-pic, .profile-map {
-    height: 210px;
+  .profile-pic,
+  .profile-map {
+    position: static;
+    float: left;
     width: 210px;
-    position: absolute;
-    top: 0;
-    left: 0;
-    overflow: hidden;
   }
 
-  .profile-pic {
-    img {
-      width: auto;
-      height: auto;
-      max-width: 210px;
-      max-height: 210px;
-    }
+  .profile-pic img {
+    width: 210px;
+    height: auto;
   }
 
   #map_canvas {
-    width: 200px;
-    height: 200px;
+    width: 210px;
+    height: 210px;
   }
 
   .small-container {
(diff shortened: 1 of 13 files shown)


mhl:
I think one of the other bits of feedback from PMG, if I remember right, is that the title of the person (e.g. "Member of the National Assembly") should be right under their name, and it isn't here, for example: http://za-pombola.staging.mysociety.org/person/beverley-lynette-abrahams/

(I'm happy to help with finding the right data to put there, it might not be obvious from the existing template.)

zarino:
@mhl – I was confused about this too. But reading my notes, it appears we agreed to just have the person's location:

```
3) Right underneath the MP's name, also include party affiliation in addition 
   to either 'Member at National Assembly” or “National Council of Provinces” 
   or “Western Cape Provincial Legislature, Eastern Cape Provincial Legislature” etc…

  - So, it currently says the party already???
  - G says she misses knowing which legislature the person is in. They have 10 
    legislatures in South Africa (rather than 2 in the UK) so it needs to make 
    it clear which legislature the person is part of.
  - Action:
    - Put ONLY the legislature under the person’s name. Everything else goes 
      in the sidebar.
```

mhl:
We may be talking at cross-purposes: those notes (in particular the "Action" bit) reads to me as if we agreed just to have an indication of the legislature under the person's name, e.g. "Member of the National Assembly" or "Member of the Western Cape Provincial Legistlature", not the location

zarino:
Oh you're right. Hmm. I wonder why I thought it was location.

So yes, if you could replace the region with the job title, that'd be ace.

mhl:
@zarino I've changed it to display important positions under the name instead - you can see that on the Pombola ZA staging site, or here are some screenshots. This could probably do with some tweaking:

one-position-title

two-position-titles

mhl:
(Otherwise :+1: thought I'll do some squashing before merging it.)

zarino:
@mhl Is that because it's too hard to just get the legislature out?

zarino:
I'm a bit worried that things like "Deputy Minister of Small Business development: National Executive (Executive) Member: National Assembly (Parliament)" is just way too much jargon. As Paul and Dave W discovered in the user testing, the bit that really matters to a layperson, to just put an MP into context, is which legislature they're part of (eg: I guess, in this case, "National Assembly").

mhl:
@zarino Sure, I've done that so you just get the name of the organisation that each person's a member of (deployed to staging too):

one-title-just-organisation-name

two-titles-just-name

Better? The "national executive" one maybe isn't so satisfactory, since you might want the post in the executive too then.

zarino:
Aaaaah :-) that's so much easier to grasp.

If anything, I might concatenate the legislatures, when there is more than one. So "National Executive, and National Assembly"?

I'm not sure how the list of organisations is selected, but we need to make sure that we're only showing important ones. Eg: here in the UK, I'd probably say "House of Commons" is important, but ["Statutory Instruments Select Committee"](http://www.theyworkforyou.com/mp/10602/david_tredinnick/bosworth#profile) isn't).

> The "national executive" one maybe isn't so satisfactory, since you might want the post in the executive too then.

I assume all this information is in the "Positions held" tab anyway?

mhl:
Yeah, the "Positions Held" tab should show all information about positions they hold or have held. The ones that might appear directly under their name are restricted by the [`important_organisations` class attribute](https://github.com/mysociety/pombola/blob/fc34c641688d0f87987297f8e47c5fdf9e26334c/pombola/south_africa/views.py#L477) so at the moment it can only be "National Assembly", "National Council of Provinces" or "National Executive".

zarino:
Great! I think with some sort of concatenation so the organisations appear on a single line rather than two lines, this will be good to go.

mhl:
OK, that last commit joins them with ' and ':

and-joining

I'll do some squashing and deploy it.

Issue #1763, opened 2015-09-14
Integrating MP attendance data into pa.org.za


paullenz:
From the C4SA API - e.g.:
You can now get per-member attendance via the API, most recent meeting first: eg. https://api.pmg.org.za/member/587/attendance/

PR #1810, opened 2015-10-20
1763 za attendance data

duncanparkes:
Add attendance data to the ZA person pages.

'Attendance' is added as a new tab, consisting of two things:

 - a sentence for each year saying what the member's committee meeting attendance has been like in that year.
 - links to the last 5 committee meetings this person attended.

This is done by us querying the PMG API serverside when someone requests a person page. In order to avoid repeating these backend queries, the results are cached for 24 hours.


duncanparkes:
Fixes #1763

duncanparkes:
(this could, of course, take a bit of @zarino time)

duncanparkes:
attendance_tab

zarino:
Yes I think we can do something with that. @duncanparkes I'll add it to the current design sprint, and see whether some space opens up for me to squeeze it in.

mhl:
I've left a few comments about very local stuff, probably too tired to think about the bigger picture clearly at the moment, but it seems fine to me :+1:

zarino:
@duncanparkes I've added some styling to make the overall percentage figures stand out, and to keep the meeting link line lengths legible.

screen shot 2015-10-22 at 10 17 29

Because this branch was started before c7f9f14 was merged into master, I had to over-specify some of the styles, to trump the `#page` style further up. When you're merging this, you should be able to remove the `#page #attendance` bits of those two styles (see comment in `_south-africa_overrides.scss:1948`) and it'll all work. (It'll work even if you *don't* remove the `#page #attendance` bits, but you *should* anyway).

duncanparkes:
@mhl I believe I've addressed all your comments here, and I've got got @zarino's styling in. I've currently lost a commit you made to remove some unneeded imports. Sorry about that - please add it back on if you can find it!

mhl:
:+1: Looks good, thanks for your work on this - we should think about https://github.com/mysociety/pombola/issues/1818 for the future.

PR #1820, opened 2015-11-06
Use jQuery UI for person page tabs

### LINKED ISSUE #1745: upgrade jQuery UI to the latest version, and use its Tabs functionality instead of simpleTabs
The aim here is to fix some bugs with the tabs on the person detail pages for South Africa, including these old ones:

* https://github.com/mysociety/pombola/issues/1292
* https://github.com/mysociety/pombola/issues/930

Currently we use [SimpleTabs](http://fotis.co/projects/simpletabs/) for these, which has had "tab selection based on URL fragment" listed as a forthcoming feature for some time. With jQuery UI Tabs, on the other hand, this is very simple, though - it already switches tab based on fragment and you can set the fragment on activating a tab with: https://github.com/mysociety/pombola/commit/b147d994a901a8ecc4023d589acb4545c1e053d8#diff-1d42a49248cf443bc27e522625fb7a6dR3

Also, I suspect the jQuery UI Tabs code is going to be better supported in the future, being very widely used.

I've got a topic branch that updates jQuery UI, makes sure that the datepicker and autocomplete still work, and switches to using the jQuery UI Tabs support, which is here:

* https://github.com/mysociety/pombola/commits/jquery-upgrade-and-use-for-tabs

The issue with this is the styling - jQuery UI comes with CSS for its widgets, using [these classes](https://api.jqueryui.com/theming/css-framework/). The bundled theme looks like:

jquery-ui

But it should look like:

old-tabs

If we remove the bundled jQuery UI CSS from the repo, it looks pretty much as before, although that has knock-on effects on the other jQuery UI widgets (e.g. the autocomplete ends up with a transparent background).

I discussed with @wrightmartin whether it would be better to remove the bundled CSS entirely, and put  back enough CSS to make the datepicker and autocomplete look right again, but he was more in favour of doing a more complete jQuery UI theme based on their standard classes for their widgets and elements - this might also be useful experience for other projects that might use jQuery UI.

So, the remaining work on this ticket is design, I think.

zarino:
Fixes #1745.

ZA person page now uses jQuery UI tabs (since it was already being used for the search autocomplete and advanced search date inputs).

jQuery doesn't come in a Sass version, so I converted the official CSS download into SCSS. I've converted some of the common colours into variables, where it helps us override them automatically based on the parent theme's settings (eg: using $colour_primary for selected dates in the date picker).

The custom bits I added to the jQuery UI styles mostly just give everything more padding, and remove lots of jQuery UI's opinionated defaults. You can see how the theme affects the jQuery UI widgets we use on the `/ui-test` page.

screen shot 2015-11-06 at 15 58 49

The default tabs style is overridden inside `south-africa_overrides.scss` to match what was there already.

There is a noticeable delay between the page loading and the jQuery UI tabs kicking in. This wasn't an issue before, but now I'm hooking all of the ZA person page tab styles off jQuery UI classes, you basically get about half a second of completely unstyled list items before jQuery UI kicks in. The quick fix is to just base all the styles off classes we control (eg: `.tab-links`) but that'll result in lots of duplication of jQuery UI styles. The *proper* solution, I fear, is to get jQuery UI running sooner after (or should I say, during) pageload.


mhl:
Do you see the border around the tab header if you click on it? I get something like this:

active-tab-border

zarino:
> Do you see the border around the tab header if you click on it?

Odd, I don't get that at all. Which browser were you using? Could the browser have cached the old styles?

zarino:
> Do you see the border around the tab header if you click on it?

Ah, actually, I *do* get it, but for me it's the buzzy blue glow that focussed elements get on a Mac.

zarino:
@mhl I've amended the commit history as you suggested, and removed the focus style on those tab links.

Obviously there's still the FOUC issue to resolve. Either we duplicate the styles, or we get jQuery running sooner. Any preference?

mhl:
Hi @zarino - thanks for those changes. If we went the route of duplicating the styles, would the duplication really be that bad? I'd have thought (*waves hands wildly*) you could use Sass mixins to minimize that.

zarino:
> would the duplication really be that bad? I'd have thought (waves hands wildly) you could use Sass mixins to minimize that.

Maybe. I’ll investigate.

zarino:
I've done some work to avoid the FOUC, but without completely hiding the background tabs in the eventuality that JavaScript doesn't run. Here's how it looks, in realtime:

fouc

mhl:
Looks good :+1: - the test failure was an issue I've fixed in master, so I've rebased and force-pushed and I'll just wait for Travis to come back green.

PR #1844, opened 2015-11-26
Change 'Search' to 'Update Results' if there is a query.

### LINKED ISSUE #1834: Mzalendo / pa.org.za - add "update results" button to search
Currently the advanced search results (see #1833) offers users the ability to filter the results based upon radio button selection or date range.  To at least this user it feels unintuitive that do to this you have to click on "Search" again and suggest that an "Update results" button is added to the advanced search function - it would have the same action as clicking search again, just be more intuitive for the user

screen shot 2015-11-26 at 16 50 42

duncanparkes:
The idea here is to make it more obvious that if you change
the things being searched for you have to press the button.

Fixes #1834.


duncanparkes:
We have an alternative plan for this.

PR #1845, opened 2015-11-26
"Update results" button on advanced search page

### LINKED ISSUE #1834: Mzalendo / pa.org.za - add "update results" button to search

zarino:
Fixes #1834.

Adds an "Update results" button to the advanced search form on all Pombolae, and improves the layout on People’s Assembly in particular by folding all the inputs onto just two lines in desktop view.

Here's a before and after:

before-and-after

The disabled "Update results" button is reactivated when any of the advanced inputs are changed:

screen shot 2015-11-26 at 16 48 23

However, because the button is disabled by javascript, there is a noticeable period on pageload where it is bright red. Any ideas @mhl?

### DIFF
--- a/pombola/core/static/sass/_search.scss
+++ b/pombola/core/static/sass/_search.scss
@@ -125,17 +125,29 @@
   padding: 1em;
   background-color: $colour_lighter_grey;
   margin-top: 1em;
+
   .radio-button-columns {
-    width: 48%;
-    margin-right: 1%;
-    @media only all and (min-width: 640px) {
-       width: 30%;
-       margin-right: 2%;
-    }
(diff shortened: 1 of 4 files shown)


mhl:
:+1: Nice work. I'm not bothered by the briefly-red button personally.

PR #1867, opened 2015-12-03
1518 A-Z links on organisation people lists

### LINKED ISSUE #1518: A - Z links on people pages
On people pages like [People at National Council of Provinces](http://www.pa.org.za/organisation/ncop/people/) PMG have requested a clickable list of letters from A to Z across the top of the page.

The use case is if you're looking for someone and you know their surname but don't know the exact spelling you can still find them by clicking through to the correct letter then looking for them. At the moment the only solution is to click through all pages using the pagination until you find the person you're looking for.

duncanparkes:
Don't merge this until @zarino has had a look for style.

Add an A-Z of links to limit people by the first letter of their name.

Fixes #1518.

### DIFF
--- a/pombola/south_africa/static/sass/_south-africa_overrides.scss
+++ b/pombola/south_africa/static/sass/_south-africa_overrides.scss
@@ -1562,12 +1562,32 @@ ul.pager {
 
 /* Organisation pages */
 
-.alphabet-links li {
-    display: inline;
+.alphabet-links {
+  @include unstyled-list;
+  @include clearfix;
+  font-size: 0.9em;
+}
 
-    .current-prefix {
-        color: $colour_green;
-    }
+.alphabet-links__link {
+  float: left;
+  text-transform: uppercase;
+  padding: 0.1em 0.2em;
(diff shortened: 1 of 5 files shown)


zarino:
@duncanparkes I've tidied up the styling slightly, and used the Peoples Assembly bottom-border convention to show the selected letter (just like tabs elsewhere on the site). It now looks like:

screen shot 2015-12-03 at 16 09 37

mhl:
This works nicely. Just a couple of things to note:

* This adds the letter pagination for URLs of the form `/organisation/{slug}/people/` but one of the pages that PMG cited as being problematic for this is the one from the "MP profiles" link on the front page, which is `/position/member/parliament/?order=name`. It also doesn't help for pages like `/organisation/national-assembly/party/anc/` which is the one that you'd get to by following the links from the `People > National Assembly` menu and then clicking on the ANC link. How do you get to pages of the form `/organisation/{slug}/people/` on the site?
* I think this could be squashed down into one commit (or two if you want to keep @zarino's contribution separate).

Issue #1854, opened 2015-12-01
[ZA] Move 2014 election data from Extras item to Public Participation tab


JenMysoc:
From PMG: “2014 Election data” should be moved and featured as a side menu to ‘Elections’ under the “Public Participation” tab.

tmtmtmtm:
![screen shot 2015-12-14 at 22 22 03](https://cloud.githubusercontent.com/assets/57483/11795929/6570289e-a2b1-11e5-8f79-94369dfb7d1f.png)

![screen shot 2015-12-14 at 22 24 19](https://cloud.githubusercontent.com/assets/57483/11795943/76dd1272-a2b1-11e5-85ca-0d9862952fa0.png)

Issue #1853, opened 2015-12-01
[ZA] Replace Extras item with Hot Topics on the menu bar


JenMysoc:
"Extras” can be replaced by “Hot Topics” and the “Hot Topics” should be duplicated as a drop down menu that includes ‘Infographics’

links to #1852

duncanparkes:
Fixed in 084698ea31fa0aac5eabf34f0e5466a6298d39c2

Issue #1852, opened 2015-12-01
[ZA] Get rid of "extras" from the menu bar on PA.org.za


JenMysoc:
linked to #1850 and #1851 and #1854 

PMG want the Extras item removed from the main menu bar along with all the items on the drop down menu.

tmtmtmtm:
![screen shot 2015-12-14 at 22 22 03](https://cloud.githubusercontent.com/assets/57483/11795899/29ce33d0-a2b1-11e5-94de-ad32cb99c49f.png)

duncanparkes:
Fixed in084698ea31fa0aac5eabf34f0e5466a6298d39c2

Issue #1851, opened 2015-12-01
[ZA] Move "links" and "Get Data" from Extras menu link to the bottom of the page


JenMysoc:
Take these two items - links and Get data - off the Extras drop down menu and put them in the footer of the page

tmtmtmtm:
![screen shot 2015-12-14 at 22 22 03](https://cloud.githubusercontent.com/assets/57483/11795913/45fd2d86-a2b1-11e5-86a5-dd17334e9abe.png)

duncanparkes:
Fixed in  7f59264070815a638d855f55f4e3797ee31debe5

PR #1876, opened 2015-12-16
1852 za extras menu

duncanparkes:
These are the code changes required for a whole collection of issues #1850, #1851, #1852, #1853, #1854.

Most of this is very routine, the one thing I wonder whether we should be doing is the changes to the footer, where it's no longer in sync with the menus.

### DIFF
--- a/pombola/south_africa/templates/footer.html
+++ b/pombola/south_africa/templates/footer.html
@@ -16,6 +16,11 @@
         {% with include_sub_menu_entries=1 %}
             {% include 'menu_entries.html' %}
         {% endwith %}
+
+        <li><a href="{% url "info_page" slug="links" %}">Links</a></li>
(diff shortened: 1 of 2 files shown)


mhl:
How about adding a `with footer=0` or `with footer=1` to the `{% include 'menu_entries.html' %}` and having a conditional based on `footer` in the `menu_entries.html` partial? (Actually, I'm not sure why the existing code was doing:

```python
         {% with include_sub_menu_entries=1 %}
             {% include 'menu_entries.html' %}
         {% endwith %}
```

... instead of:

```
        {% include 'menu_entries.html' with include_sub_menu_entries=1 %}

```

... for that matter. Maybe that wasn't in Django [back in 2012](https://github.com/mysociety/pombola/commit/197a5c23778729).)

PR #1880, opened 2015-12-17
1857 more positions in tagline

### LINKED ISSUE #1856: [ZA] Ability to reflect if an MP is a "former MP"
From PMG: "We would like it to reflect on the MP profile if an MP is an “Former MP” - this is not happening: http://www.pa.org.za/person/graham-peter-dalziel-mac-kenzie/"

### LINKED ISSUE #1857: [ZA] Allow for label of MLP under person name on person page
From PMG: "We decided that Member at “National Assembly”, “National Council of Provinces” or “Western Cape Provincial Legislature” should be captured underneath MP’s name. Thanks for doing that, but we noticed that it has not been done for Members of Provincial Legislatures (MPLs)"

We need to add this option in the same way we added the other options

### LINKED ISSUE #1878: Dedupe organisations displayed under a person's name
Note that National Council of Provinces appears twice here!

http://www.pa.org.za/person/dr-mateme-hunadi-euphemia/

2015-12-17-113552_1366x768_scrot

duncanparkes:
Add memberships of Provincial Legislatures to the tagline on a person page (closes #1857). This is done by changing how we decide which positions are important to work off the slug of the organization kind rather than off the organization. In future we could make this piece of configuration data by having a table of ids of org kinds which are this sort of important in the database and editable by the Django Admin.

Also adds former memberships of important organizations (closes #1856).

Finally, I've removed some of odd bits of code which weren't used anywhere else.

I suggest we don't actually put this live until we've added some better testing of the person detail page and also fixed the problem in #1878.

### DIFF
--- a/pombola/core/models.py
+++ b/pombola/core/models.py
@@ -807,11 +807,6 @@ class Place(ModelBase, ScorecardMixin, BudgetsMixin):
         positions = self.all_related_positions().current_politician_positions()
         return Person.objects.filter(position__in=positions).distinct()
 
-    def all_related_former_politicians(self):
-        """Return a query set of all the former politicians for this place, and all parent places."""
-        positions = self.all_related_positions().former_politician_positions()
-        return Person.objects.filter(position__in=positions).distinct()
-
     def child_places_by_kind(self):
         """Return all concurrent child places, grouped by their PlaceKind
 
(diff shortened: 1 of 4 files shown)


tmtmtmtm:
As the problem in #1878 is pre-existing, I don't think this should be blocked on solving that. If it's easy to fix up some things like that _en passant_, then we should certainly do so, but otherwise we should prioritise them separately.

mhl:
:+1: Looks good to me...

PR #1881, opened 2015-12-17
Flake8ify

duncanparkes:
This is the very start of what I expect to be a fairly long running task to make the whole of the codebase PEP8 compliant. In this PR I've just run flake8 with the tests for unneeded imports etc. We should add a check for each PEP8 test once we have it passing and make that part of the continuous integration so that we never go backwards.

There are two sets of changes here which I think are worth considering.

1) Does `from fabric.api import *` do any magic? We don't appear to use anything that's imported that way.

2) Should we just put `# flake8: noqa` on `pombola/ghana/management/feedparser.py`, which is vendored? Better still, can we include it from PyPI instead (#1882)

### DIFF
--- a/bin/deskew-and-unpaper.py
+++ b/bin/deskew-and-unpaper.py
@@ -7,7 +7,7 @@
 # "convert -deskew '40%'" and then "unpaper" to remove scanning
 # artefacts.  (ImageMagick does better than unpaper on large skews.)
 
-import os, re, sys
+import os, re
 from subprocess import check_call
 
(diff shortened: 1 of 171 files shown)


dracos:
"1) Does from fabric.api import * do any magic?” - aren’t the sudo/ctl functions in e.g. fab/nginx.py coming from that import?

mhl:
These should all be squashed together, I think

mhl:
OK, I think I've been through everything, and made some small comments - looks fine other than those.

PR #1894, opened 2016-01-04
ZA: Social media links in header

### LINKED ISSUE #1849: Move social media icons to the top of the homepage
From PMG: "We also wanted the Facebook and Twitter icons moved to the top so that its immediately visible on the home page"

This is their second highest priority action.

zarino:
Fixes #1849.

Also includes a few overdue fixes to the horizontal padding / gutter either side of the header elements on mid-size screens, and the overcomplicated layout that the "about-this-site" menu items used to have.

Large screen:

screen shot 2016-01-04 at 12 44 22

Mid-size screen: (this is at its absolute worst – 640px)

screen shot 2016-01-04 at 12 45 54

Small screen:

screen shot 2016-01-04 at 12 46 29

### DIFF
--- a/pombola/south_africa/static/sass/_south-africa_overrides.scss
+++ b/pombola/south_africa/static/sass/_south-africa_overrides.scss
@@ -333,10 +333,11 @@ $logo-width-large: 184px;
     height: auto;
     background-image: none;
     background: $colour_main_background;
-    padding: .5em 0 0;
+    padding: 0.5em 0 0 0;
     position: relative;
+
     @media only all and (min-width: 640px), print {
-        padding: 0;
+        padding: 0 0.5em 0 0.5em; // stop children touching edges of window on narrow desktops
         height: 114px;
     }
 }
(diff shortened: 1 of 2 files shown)


pombola/south_africa/static/sass/_south-africa_overrides.scss
```diff
@@ -555,6 +573,32 @@ menu#mobile-top-tools,
     }
 }
 
+@media only all and (min-width: 640px) {
+    .about-this-site-menu__facebook,
+    .about-this-site-menu__twitter {
+        position: absolute;
+        top: (114px / 2) - (38px / 2);
```
mhl: Thanks for explaining this in IRC - it'd be great if you could add a comment explaining where the 114px and 38px values come from, and that this is vertically centring the icon in the header...

mhl:
I just added a comment asking for a comment, otherwise looks fine (and thanks for the bonus commit to fix the margins in the header). :+1:

PR #1901, opened 2016-01-11
1886 ncop grouped as other

### LINKED ISSUE #1886: ZA: NCOP members wrongly grouped as 'Other' on Place pages
http://www.pa.org.za/organisation/ncop/people/ shows all the National Council of Provinces members:

screen shot 2015-12-19 at 12 37 21

However, when you click on one of those province names to go to the Area page (e.g. http://www.pa.org.za/place/mpumalanga/), this lists 0 members of the NCOP:

screen shot 2015-12-19 at 12 44 12

with the actual members being grouped under 'Other' instead:

screen shot 2015-12-19 at 12 45 00

duncanparkes:
This sorts out NCOP delegates not appearing in the right place, which was caused by the position having title 'delegate' rather than member. The code has loads of places which make allowance for NCOP positions being delegates rather than members, so I don't really want to just switch it to member as it might well break some.

It would be nice to generally reduce the number of places the code refers to particular slugs.

At the moment this causes another test to fail. There is a check to make sure that if someone is both a member of the National Assembly and a delegate to NCOP that they don't appear on the list for the place twice. I would like to know if this is a real possibility, and if not, that test could be cut.

Fixes #1886.

2016-01-11-162742_1366x768_scrot

### DIFF
--- a/pombola/south_africa/tests.py
+++ b/pombola/south_africa/tests.py
@@ -1481,6 +1481,11 @@ class SAPlaceDetailViewTest(WebTest):
             slug='member',
         )
 
+        self.positiontitle_delegate = models.PositionTitle.objects.create(
+            name='Delegate',
+            slug='delegate',
+        )
+
         self.positiontitle_other = models.PositionTitle.objects.create(
             name='Other',
             slug='other',
@@ -1514,7 +1519,7 @@ class SAPlaceDetailViewTest(WebTest):
         models.Position.objects.create(
             organisation=self.org_ncop,
             place=self.place,
-            title=self.positiontitle_member,
+            title=self.positiontitle_delegate,
             person=person_related_ncop,
             category='political',
             end_date='future',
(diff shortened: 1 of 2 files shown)


tmtmtmtm:
@JenMysoc can you check whether it's possible for someone to be simultaneously a member of the National Assembly and the NCOP? We suspect they probably can't, but it would be worth making sure.

tmtmtmtm:
@duncanparkes removing that test doesn't seem to have made Travis happy…

mhl:
The issue with an unrelated test failing was because these test methods weren't isolated because the base class was `TransactionWebTest` rather than `WebTest`. I've pushed a commit to master that removes erroneous uses of `TransactionWebTest`, which is 1b02106222 - the commit message explains that further.

I've rebased this branch over the new master and added a fixup for another test that depended on having the right PositionTitle for NCOP delegates. The tests then all pass without having to remove the test that apparently appeared problematic (because of bad test isolation, I assume).

The only comment I'd add is that I don't think it's worth having separate commit with the tests in a failing state, so I'd squash them out. Otherwise this looks good.

duncanparkes:
2016-01-19-144740_1366x768_scrot

Issue #1869, opened 2015-12-09
Make a public copy of the database available


chrismytton:
With the auth info removed. This would make it easier to get up and running. It would also be nice if this could work with the `Vagrantfile` so that you can get a full working dev environment with one command.

akuckartz:
:+1:

tmtmtmtm:
Closed by https://github.com/mysociety/pombola/pull/1906

PR #1906, opened 2016-01-15
crontab: Do a pg_dump of selected tables from the database every day

mhl:
It's useful for getting a developer up-and-running with Pombola to
have a database dump available; it would also be useful as part of
the provisioning of a Vagrant box to be able to restore a database
dump to get a realistic site.  This commit gets cron to do a pg_dump
of the database daily which can be used in both these situations.

The risk here is that there is personal information in the database
(e.g. the auth_user table, the django_session table, feedback
submitted to the site, etc.)  To avoid this we specify only the
tables we want to dump, and have looked through the data in these
by hand.  There's still room for error here, but it's safer than
trying to exclude every table that might include personal data.

### DIFF
--- a/conf/crontab.ugly
+++ b/conf/crontab.ugly
@@ -22,6 +22,18 @@ PYTHONIOENCODING=utf-8
 # check that no bad slugs have been stored in the database
 0 23 * * * !!(*= $user *)!! run_management_command core_list_malformed_slugs
 
+!!(*
+    %dump_times = (
+        'mzalendo.mysociety.org' => 10,
+        'www.pa.org.za' => 20,
+        'www.shineyoureye.org' => 30,
+        'www.kuvakazim.com' => 40,
+    );
+    if (exists $dump_times{$vhost}) {
+        $dump_directory = "/data/vhost/$vhost/media_root/dumps";
+*)!!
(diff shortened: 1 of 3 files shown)


chrismytton:
Would this be suitable for making public? https://github.com/mysociety/pombola/issues/1869

mhl:
@chrismytton Thanks for linking the issues. The intention is to make it public, yes (it would be after merging and deploying this PR) - I think most of the questions with this PR are about the risks of doing that.

mhl:
(Sorry, a bit of confusion there about whether I was commenting on the issue or pull request :weary:)

tmtmtmtm:
How do we ensure that this is always kept up to date when new tables are added?

mhl:
@tmtmtmtm The new version should address this - it'll error (and the output will be emailed) if any unknown tables are found.

tmtmtmtm:
Great!

duncanparkes:
This all looks good to me. Glad to see the `votematch_submission` being added to ignored ones in the squash commit. :+1:

tmtmtmtm:
The 'files changed' tab here is now empty, but this change can be seen at https://github.com/mysociety/pombola/commit/c0a23138aa33779fc2c4a917e530b7f0fd960f14

PR #1915, opened 2016-01-19
ZA: Add capability to show Facebook, LinkedIn etc on person page

### LINKED ISSUE #1316: ZA: Profile pages should show all contact information
Only certain contact information is currently displayed on the site. E.g. for http://www.pa.org.za/person/pieter-van-dalen/ we have YouTube, LinkedIn, and Who's Who data but this does not appear.

There also seems to be a bit of uncertainty around use of the telephone contact kinds - e.g. phone, cell, and voice (this MP is one example - we have the same number listed for both phone and cell).

zarino:
Fixes #1316.

Displays Facebook, LinkedIn, YouTube, and Who’s Who links for people, if we have them in the database.

Since Facebook, LinkedIn, and YouTube URLs are either always, or very often, just random IDs (rather than nicely formatted usernames, like on Twitter), I've decided to just show the contact service name, rather than giving a preview of the user’s name on that service.

This *does* mean, if someone has two Facebook pages (eg: a personal page and a party candidate page), then they'll both just be displayed as "Facebook". Is that a problem?

The row of buttons looks pretty busy when all six are active, but I'm under the impression that will be a fairly uncommon happening. Is that correct?

screen shot 2016-01-19 at 14 39 00


zarino:
Worth noting, although this *technically* fixes #1316, there is still some education and/or data cleaning required, to avoid the duplication of phone numbers also reported in the original ticket – and indeed, the malformation of a lot of the existing Facebook/LinkedIn/YouTube etc URLs in the database.

duncanparkes:
:+1:

duncanparkes:
To tidy the twitter handles, in `python manage.py shell`.

```
>>> from pombola.core.models import Contact
>>> for contact in Contact.objects.filter(kind__slug='twitter'):
...     contact.value = contact.value.strip().strip('@')
...     contact.save()
... 
>>> 
```

PR #1925, opened 2016-01-27
Changing search suggestions

### LINKED ISSUE #1757: Replace "Goodluck Jonathan" as suggested search term
The search box suggests I search for Goodluck Jonathan. Doing so returns zero results. Searching solely for Jonathan gives me several results (though none of them him), but then tells me that it found nothing:

screen shot 2015-09-03 at 16 01 32

JenMysoc:
fixes #1757


tmtmtmtm:
We don't really have good results for this search either, though:  http://www.shineyoureye.org/search/?q=Muhammadu+Buhari

Perhaps there's someone better from the legislative branch, rather than the president, who doesn't seem to have much to do with Parliament?

JenMysoc:
Ok have changed to Babatunde Fashola instead as he has a person page and appears to be a well liked politician.

PR #1926, opened 2016-01-27
Mp attendance page

xybrnet:
This branch implements a page to show MP attendance data obtained from api.pmg.org.za.

Data is retrieved from the API, manipulated in the view and presented in a table on it own page. There is a link to this page from the home page. (Retrieving data by year is currently not supported by the API at the moment.)

The Datatables jQuery library is used to make the table sortable by columns, and for client side search.
The external assets for this is only loaded on the MP attendance page.

### DIFF
--- a/pombola/south_africa/templates/south_africa/mp_attendance.html
+++ b/pombola/south_africa/templates/south_africa/mp_attendance.html
@@ -1,0 +1,110 @@
+{% extends 'base.html' %}
+{% load compressed static %}
+
+{% block title %}MP Attendance{% endblock %}
+
+{% block extra_js_to_load %}
+    addExtraJS([ {% compressed_js 'attendance-table' %} null ]);
+{% endblock %}
+
+{% block css_headers %}
+  {{ block.super }}
+  {% compressed_css 'datatables' %}
+{% endblock %}
+
+{% block body_attributes %} class="mp-attendance" {% endblock %}
+
+{% block content %}
+<h1 class="page-title">MP Attendance</h1>
+<div class="major-column">
+  <p>Explore the commitee meeting attendance records for the members of parliament.</p>
+</div>
+
+<div class='mp-attendance-filter'>
+  <div class="col-50">
+    <form class='select-year' action="{% url 'mp-attendance' %}" method="get">
+      <div class="filter-option">
+        <label>Select a year:</label>
+        <select name="year">
+         {% for yr in years %}
+          <option value="{{ yr }}" {% if yr == year %}selected{% endif %}>{{ yr }}</option>
+         {% endfor %}
+        </select>
+      </div>
+      <div class="filter-option">
+        <label>Choose a party:</label>
+        <select name="party">
+          <option value="" selected>All</option>
+         {% for pty in parties %}
+          <option value="{{ pty }}" {% if pty == party %}selected{% endif %}>{{ pty }}</option>
+         {% endfor %}
+        </select>
+      </div>
+      <input class="button" type="submit" value="Show attendance">
+    </form>
+  </div>
+
+  {% if party %}
+   <div class="col-50">
+    <h2>{{ aggregate_attendance }}%</h2>
+    <p>Average attendance for {{ party }}</p>
+   </div>
+  {% endif %}
+
+  <div class="clear"></div>
+
+  <hr>
+  <a class="button download" href="{{ download_url }}">Download raw data</a>
+
+</div>
+
+<table id="mp-attendance">
+  <thead>
+    <tr>
+      <th>Name</th>
+      <th>Party</th>
+      <th>Meetings</th>
+      <th>Present</th>
+      <th>Arrived late</th>
+      <th>Departed Early</th>
+    </tr>
+  </thead>
+  <tbody>
+   {% for member in attendance_data %}
+    <tr>
+      <td><a href="{{ member.pa_url }}">{{ member.name }}</td>
+      <td>{{ member.party_name }}</td>
+      <td>{{ member.total }}</td>
+      <td width="15%">
+        <div class="attendance">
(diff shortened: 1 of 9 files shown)


pombola/south_africa/templates/south_africa/mp_attendance.html
```diff
@@ -0,0 +1,112 @@
...
+{% block content %}
+<h1 class="page-title">MP Attendance</h1>
+<div class="major-column">
+  <p>Explore the commitee meeting attendance records for the members of parliament.</p>
+</div>
+
+<div class='mp-attendance-filter'>
+  <div class="col-50">
+    <form class='select-year' action="{% url 'mp-attendance' %}" method="get">
+      {% csrf_token %}
```
mhl: The `{% csrf_token %}` is unnecessary since all the requests are made with `GET` now, and leaving this in means you end up with the CSRF token in the URL when changing the view of the data.

pombola/south_africa/templates/south_africa/mp_attendance.html
```diff
@@ -0,0 +1,112 @@
...
+   <div class="col-50">
+    <h2>{{ aggregate_attendance }}%</h2>
+    <p>Average attendance for {{ party }}</p>
+   </div>
+  {% endif %}
+
+  <div class="clear"></div>
+
+  <hr>
+    <a href="{{ download_url }}">
```
mhl: I think this could be simplified to:

    <a class="button download" href="{{ download_url }}">Download raw data</a>

pombola/south_africa/views.py
```diff
@@ -1962,3 +1962,101 @@ def get_context_data(self, **kwargs):
...
+
+    def calculate_abs_percenatge(self, num, total):
+        """
+        Return the truncated value of num, as a percentage of total.
+        """
+        return int("%.*f" % (0, (num / total * 100)))
+
+    # def post(self, request, *args, **kwargs):
+    #     context = self.get_context_data()
+    #     return super(SAMpAttendanceView, self).render_to_response(context)
```
mhl: Could you possibly remove the commented out `post` method?

pombola/south_africa/views.py
```diff
@@ -1962,3 +1962,101 @@ def get_context_data(self, **kwargs):
             context['data'] = None
 
         return context
+
+class SAMpAttendanceView(TemplateView):
```
mhl: I think the pages served from this view could be cached safely - perhaps using cache_control to set the max age to 10 minutes would be worthwhile?

pombola/south_africa/views.py
```diff
@@ -1962,3 +1962,101 @@ def get_context_data(self, **kwargs):
...
+
+        for annual_attendance in data['results']:
+            year = str(dateutil.parser.parse(annual_attendance['end_date']).year)
+            context['years'].append(year)
+
+            if year == context['year']:
+                parties = set(ma['member']['party_name'] for
+                    ma in annual_attendance['attendance_summary'])
+                parties.discard(None)
+                context['parties'] = sorted([p for p in parties])
```
mhl: This can just be `sorted(parties)`

pombola/south_africa/views.py
```diff
@@ -1962,3 +1962,101 @@ def get_context_data(self, **kwargs):
...
+
+                attendance_summary = annual_attendance['attendance_summary']
+                if context['party']:
+                    attendance_summary = [mbr_att for mbr_att in attendance_summary if
+                        mbr_att['member']['party_name'] == context['party']]
+
+                aggregate_total = aggregate_present = 0
+
+                for summary in attendance_summary:
+                    total = sum([v for v in summary['attendance'].itervalues()])
```
mhl: It's a tiny and nearly insignificant thing here, but these could just be generator expressions rather than list comprehensions that are passed to `sum` to save on constructing a list, e.g.

    sum(v for v in summary['attendance'].itervalues())

pombola/south_africa/views.py
```diff
@@ -1962,3 +1962,101 @@ def get_context_data(self, **kwargs):
...
         return context
+
+class SAMpAttendanceView(TemplateView):
+    template_name = "south_africa/mp_attendance.html"
+
+    def calculate_abs_percenatge(self, num, total):
+        """
+        Return the truncated value of num, as a percentage of total.
+        """
+        return int("%.*f" % (0, (num / total * 100)))
```
mhl: I don't think we've documented this anywhere, but we've generally been trying to use `.format` instead of `%` for string interpolation in Pombola

mhl:
This looks really good, thanks!  I've left a few comments inline, but I think they're all minor things.  A couple of other things that might be worth mentioning:

* I think this could be squashed into two commits without losing value to people looking back at the history - I was thinking one to introduce the external datatables library (ideally with a commit message mentioning the source URL, the version and its license) and then another commit for everything else. (BTW, I noticed one of these commits had a subject line that was too long - we [try to keep them under 72 characters](http://mysociety.github.io/coding-standards.html#commit-messages).)
* It's great you've thought about the responsive design for this - one small oddity, though, is that if you scroll to the right of the table on small screen sizes, there's a dark background at the right of some of the blocks:

mp-attendance

I hope these comments are of some use

xybrnet:
Thank you for the quick feedback. I've made all the changes requested, except for the responsive mobile view.

I'm trying to get the table to behave in a similar manner to Bootstrap responsive tables, but I'm unable to do so due the page-wrapper class having the following property: `display: table-cell`.

I can look into this some more, though I'm not able to work on it in the next few days, as I'm tied up at an event. Could this be merged so long?

mhl:
Thanks! I've merged and deployed this now - I created a couple of issues for two small followups so anyone who has time could look at them:

* https://github.com/mysociety/pombola/issues/1930
* https://github.com/mysociety/pombola/issues/1931

Issue #1838, opened 2015-11-20
Constituency box should only show constituency of Parliamentary Memberships


tmtmtmtm:
If other memberships are (probably erroneously) set to have a constituency, these end up displayed in the Constituency box on a member page (see #1831)

@paullenz tidied up the incorrect data for the specific cases we could find, but we should also ensure that this box _only_ looks at the constituency for memberships of the legislature itself, so that even if bad data is entered in future, this error will not reappear.

tmtmtmtm:
This is ticketed against ZA, as we know it's definitely an issue there, but we should also ensure that this doesn't happen on other sites too.

tmtmtmtm:
@paullenz this is still a problem for 
* http://info.mzalendo.com/person/naomi-shaban/
* http://info.mzalendo.com/person/maison-leshoomo/
* http://info.mzalendo.com/person/mutava-musyimi/
* http://info.mzalendo.com/person/moses-ole-sakuda/

paullenz:
Looking at e.g. 

![screenshot 2015-11-23 11 12 14](https://cloud.githubusercontent.com/assets/909811/11335298/1bda22fa-91d3-11e5-9d50-649d79ea26a3.png)

I can only see a single constituency  listed?

paullenz:
(http://info.mzalendo.com/person/mutava-musyimi/)

tmtmtmtm:
@paullenz I don't really understand the conditions under which a "County" appears in this box, but this is triggered here by him having "Member of The National Alliance (TNA) for Embu" set, which is the same underlying problem.

paullenz:
It will be triggered by that, yes - but is the county display an issue per
se?  It is not a dupe constituency

tmtmtmtm:
It depends slightly on what the purpose of this field is — is there something that explains it somewhere? 

It's not necessarily a problem to show both, but if it _should_ be shown here, it would be good if it was for the right reason, rather than the wrong one.

tmtmtmtm:
Today's update caused problems due to these ones having multiple constituencies:
* http://info.mzalendo.com/person/simon-ogari/
* http://info.mzalendo.com/person/mwalimu-mwahima/
* http://info.mzalendo.com/person/maison-leshoomo/
* http://info.mzalendo.com/person/mohamed-elmi/

And http://info.mzalendo.com/person/naomi-shaban/ have both a county and a constituency

[...]

PR #1928, opened 2016-01-29
1838 important orgs

duncanparkes:
These changes should hopefully stop duplicate constituencies/counties appearing on the sidebar on the person detail page (#1838). There should now be at most one constituency for NA members, and one county for senators.

This is currently done in the view and template layers for KE only. It would be good in the long run to have a system for storing which combinations of organisation, organisationkind, positiontitle, etc should be considered important enough to display here, and storing that in the database rather than the code.

### DIFF
--- a/pombola/kenya/views.py
+++ b/pombola/kenya/views.py
@@ -1,6 +1,7 @@
 # Create your views here.
 # -*- coding: utf-8 -*-
 
+import logging
 import random
 import json
 import sys
@@ -17,7 +18,7 @@ from .forms import (
     YouthEmploymentInputForm
 )
 
-from pombola.core.models import Person
+from pombola.core.models import Person, Place
 from pombola.core.views import PersonDetail, PersonDetailSub
 from pombola.experiments.views import (
     ExperimentViewDataMixin, ExperimentFormSubmissionMixin,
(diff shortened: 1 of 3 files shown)


pombola/kenya/views.py
```diff
@@ -139,6 +139,19 @@ def get_context_data(self, **kwargs):
             if shujaaz.FINALISTS_DICT[year].get(self.object.pk)
         ]
 
+        political_positions = self.object.position_set.all().political().currently_active()
+
+        na_memberships = political_positions.filter(title__slug='member-national-assembly')
+        senate_memberships = political_positions.filter(title__slug='senator')
+        if na_memberships:
```
mhl: I just wonder if it'd be worth throwing an exception if there's > 1 `na_memberships` or > 1 `senate_memberships` so we'll know there's something to fix?

mhl:
One small query but otherwise this looks good, thanks! :+1:

pombola/kenya/views.py
tmtmtmtm: Would definitely be good to at least log that somewhere.

duncanparkes:
@mhl It should now email admins if there are two NA or Senate memberships for the same person at the same time, but will still display the most recent. Does this look OK?

PR #1933, opened 2016-02-03
hansard: Fix for it being possible to create aliases to hidden people

mhl:
People who have 'hidden' set to True are people who have stood for an
election but essentially zero chance of ever having appeared as a
speaker in Hansard.  Since there were Alias objects referring to hidden
people, let's prevent this from being possible.

### DIFF
--- a/pombola/core/lookups.py
+++ b/pombola/core/lookups.py
@@ -1,0 +1,11 @@
+from ajax_select import LookupChannel
+
+from .models import Person
+
+
+class PersonLookup(LookupChannel):
+
+    model = Person
+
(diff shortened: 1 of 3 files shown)


duncanparkes:
Looks like it should fix the problem, worth adding order to preserve original behaviour :+1:

mhl:
(Repeating a bit of discussion from IRC) - the results are ordered, in fact, because of the [`ordering` specified in `Person`'s `Meta` class](https://github.com/mysociety/pombola/blob/ee16ec2aae755639a614382c0e49f4e343f4d207/pombola/core/models.py#L500-L501).

Issue #1938, opened 2016-02-04
Import constituency office data and photos


duncanparkes:
PMG have been undertaking an exercise to collect accurate locations and photos of all the party constituency offices.

We need:

  - [x] A new script to import their data and update our database with the correct locations
  - [x] The script to also store copies of the photos
  - [x] To display the photos on each constituency office page [do #1948 at same time]

duncanparkes:
This possibly relates to #1311

duncanparkes:
d6bf2d6b4d53bf65b347c0c0b772271ccce75f0a Does the office name matching and location update parts of this.

zarino:
Regarding the display of the photos: Given there will be more offices *without* a photo than *with*, putting it somewhere structurally insignificant would be best (so there isn't an obvious gap when no photo is available). I'm thinking below the map in the sidebar:

![screen shot 2016-02-15 at 17 17 25](https://cloud.githubusercontent.com/assets/739624/13055580/07f31766-d408-11e5-9627-e3cf29a1ce8c.png)

The downside of this is that it pushes the address and phone details further down. Do you think that will be an issue @JenMysoc @mhl?

Do we know anything about the aspect ratio of the images? (ie: Are they all landscape?)

tmtmtmtm:
I'd lean towards even putting it _after_ the address — as I'm assuming the primary use case here is so that you can recognise it if you go visit, which would request getting the address first… (I may be completely misunderstanding what this is for, though)

zarino:
Something like this? @tmtmtmtm @JenMysoc 

![screen shot 2016-02-16 at 16 09 37](https://cloud.githubusercontent.com/assets/739624/13082225/ba421da8-d4c7-11e5-86ca-f31677a8ac24.png)

JenMysoc:
:+1: Looks awesome to me

zarino:
Here's roughly how it'll look on a mobile – with the photo about halfway down the page, after the map and contact details:

![screen shot 2016-02-16 at 16 11 00](https://cloud.githubusercontent.com/assets/739624/13082286/fac2ea6a-d4c7-11e5-89ff-36da9a6be12f.png)

zarino:
I've pushed this little bit of styling to the `1938-import-new-constituency-office-data` branch.

Now it needs a developer (@duncanparkes @mhl?) to replace the hard-coded example image with the actual image for the constituency office.

PR #1941, opened 2016-02-08
Add a script to import new constituency office locations.

duncanparkes:
PMG are conducting an exercise to go to all the constituency offices
and get accurate locations for them and photographs. The data they
collect is available as a CSV, and this new management command
processes that CSV and updates our data with the improved locations.

Fixes part of #1938.

### DIFF
--- a/pombola/south_africa/management/commands/south_africa_import_new_constituency_office_locations.py
+++ b/pombola/south_africa/management/commands/south_africa_import_new_constituency_office_locations.py
@@ -110,6 +110,16 @@ class Command(BaseCommand):
                 # Now get the photo
                 photo_url = row.get('Constituency_Photo')
 
+                try:
+                    org.images.get(source=photo_url)
+                    print (
+                        "Skipping {} as url matches existing image."
+                        .format(org.slug)
+                        )
+                    continue
+                except Image.DoesNotExist:
+                    print "Adding new image to {}.".format(org.slug)
+
                 if photo_url:
                     response = requests.get(photo_url)
                     image_filename = '{}'.format(org.slug)
--- a/pombola/south_africa/management/commands/south_africa_import_new_constituency_office_locations.py
+++ b/pombola/south_africa/management/commands/south_africa_import_new_constituency_office_locations.py
@@ -12,14 +12,25 @@ on constituency offices to have the correct location.
 
 
 import csv
+import os.path
 import re
+from StringIO import StringIO
 
+import requests
+
+from django.conf import settings
+
+from django.core.files.base import ContentFile
+from django.core.files.storage import FileSystemStorage
 from django.core.management.base import BaseCommand, CommandError
 from django.contrib.gis.geos import Point
 
 from pombola.core.models import (
     Organisation,
     )
+from pombola.core.utils import mkdir_p
+
+from pombola.images.models import Image
 
 
 # Matches the format of the locations in the CSV we get from PMG.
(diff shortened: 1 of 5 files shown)


duncanparkes:
I've added a script to store copies of the photos of the offices, and changed @zarino's template bit to use the real image.

There is a single constituency office (LadySmith, ANC) which gets two lines in the CSV file. I'm not sure what that's about, but we'll finish up with both images in the DB, one as primary.

pombola/south_africa/management/commands/south_africa_import_new_constituency_office_locations.py
```diff
@@ -0,0 +1,138 @@
...
+    Organisation,
+    )
+from pombola.core.utils import mkdir_p
+
+from pombola.images.models import Image
+
+
+# Matches the format of the locations in the CSV we get from PMG.
+location_re = re.compile(
+    r'SRID=4326;POINT\(-?([0-9]+\.[0-9]+) (-?[0-9]+\.[0-9]+)\)')
```
mhl: The first group doesn't include the `-` if it's present. (The longitude won't be negative in South Africa, but this might get reused elsewhere.)

pombola/south_africa/management/commands/south_africa_import_new_constituency_office_locations.py
```diff
@@ -0,0 +1,138 @@
...
+                qs = constituency_offices.filter(
+                    name__regex=r'^{}.+: {}$'.format(party, name))
+
+                try:
+                    org = qs.get()
+                except Organisation.MultipleObjectsReturned:
+                    print "{} Orgs returned: {}".format(
+                        qs.count(),
+                        repr(qs),
+                        )
```
mhl: I think the script should exit in this case - at the moment it'll struggle on but with the old value for `org`

pombola/south_africa/management/commands/south_africa_import_new_constituency_office_locations.py
```diff
@@ -0,0 +1,138 @@
...
+                        .format(org.slug)
+                        )
+                    continue
+                except Image.DoesNotExist:
+                    print "Adding new image to {}.".format(org.slug)
+
+                if photo_url:
+                    response = requests.get(photo_url)
+                    image_filename = '{}'.format(org.slug)
+                    image_filepath = os.path.join(
```
mhl: This will save the image directly to the media root, which potentially might overwrite an existing file. (Presumably nothing else is creating images in that subdirectory, but that's not a safe assumption.) I believe that the right way to do it, which is used in the example I pointed out to you (https://github.com/mysociety/yournextrepresentative/blob/master/candidates/models/popolo_extra.py#L581-L585), is to create a FileSystemStorage object, and call save on it with the "desired filename". It might end up writing the image to a safely found variant of that filename if it already exists.

mhl:
I've left a few comments to deal with, but otherwise this looks good to go.

pombola/images/migrations/0002_auto_20160229_1343.py
```diff
@@ -0,0 +1,21 @@
+# -*- coding: utf-8 -*-
```
mhl: Could you give this migration a more descriptive name than "auto_20160229_1343"? :)

pombola/south_africa/management/commands/south_africa_import_new_constituency_office_locations.py
```diff
@@ -0,0 +1,136 @@
...
+                    continue
+                except Image.DoesNotExist:
+                    print "Adding new image to {}.".format(org.slug)
+
+                if photo_url:
+                    response = requests.get(photo_url)
+                    image_filename = '{}'.format(org.slug)
+
+                    desired_storage_path = os.path.join(storage_path, image_filename)
+                    storage_filename = storage.save(desired_storage_path, StringIO(response.content))
```
mhl: Instead of using a StringIO, could you't you do:

```python
requests.get(photo_url, stream=True)
...
storage_filename = storage.save(desired_storage_path, response.raw)
```

PR #1944, opened 2016-02-10
Return None for admin url of missing foreign key.

duncanparkes:
(instead of a traceback).

If for some reason the object that the generic foreign key on an identifier doesn't exist, show '(None)' on the admin page for all identifiers for the link to this foreign key, rather than giving a traceback.
2016-02-10-091720_1366x768_scrot

At the moment, this error makes it hard to edit the bad data from the admin interface. I guess it might be worth sending an email to admins anyway?


mhl:
Closed in favour of #1945

PR #1951, opened 2016-02-12
Upgrade SayIt (used for ZA) to the latest version (Part 1)

mhl:
This pull request is related to multiple other pull requests.  It's part 1 of the
update to Pombola, since after this is merged and deployed, the command:

    ./manage.py pombola_sayit_sync_pombola_to_popolo

... needs to be run. Then [part 2](https://github.com/mysociety/pombola/pull/1952) can be deployed.

The other bits of related work which need to be reviewed at the same time are:

* The corresponding update to za-hansard: https://github.com/mysociety/za-hansard/pull/71
* and [everything in popolo-name-resolver since it forked from popit-resolver](https://github.com/mysociety/popolo-name-resolver/compare/a0d8fa25842fb...master?expand=1).

I would suggest reviewing this in the order:

* popolo-name-resolver
* za-hansard
* pombola part 1
* pombola part 2

Also, while I've got the tests running and passing (with some better tests for
popolo-name-resolver) I haven't yet tried a reparse of any of the existing
proceedings - it would be good to do that for all of:

* Committees
* Questions and answers
* Hansards

However, I can do that while someone's working on the reviewing of this new
code.


chrismytton:
I'm struggling to get this running locally, but from what I can tell reading the code I think [`SASpeakerRedirectView`](https://github.com/mysociety/pombola/blob/002bb3c8997d36c66a145cfc7db2e6df9d66b129/pombola/south_africa/views.py#L835-L847) needs to be updated to use the new models which connects a pombola person to a sayit speaker?

chrismytton:
I think the [South Africa docs](https://github.com/mysociety/pombola/blob/sayit-upgrade-part-1/docs/south_africa.md#reparsing-hansard) will need to be updated to reflect the changes in sayit (both the section I've linked to and possibly other parts of that doc).

chrismytton:
Quite scary how big this diff id! I've left a couple of comments, but overall this is looking good! :+1:

chrismytton:
> I'm struggling to get this running locally, but from what I can tell reading the code I think SASpeakerRedirectView needs to be updated to use the new models which connects a pombola person to a sayit speaker?

Ah, I see that's been done in part 2!

mhl:
Thanks for the review, @chrismytton - I've just deployed this, but I created a new issue for updating the docs.

PR #1952, opened 2016-02-12
Upgrade SayIt (used for ZA) to the latest version (Part 2)

mhl:
This updates the SayIt-ish templates / views to use the pombola_sayit join table.
However, this should only be deployed after deploying [Part 1](https://github.com/mysociety/pombola/pull/1951)
and running the `pombola_sayit_sync_pombola_to_popolo` command since
otherwise these updated templates won't show thumbnails of speakers and
won't link through to the speakers' pages.


chrismytton:
:+1: Looks good to me

mhl:
Merged to master now...

Issue #1914, opened 2016-01-19
[NG] Placeholder for making a "pombolaLite" for SYE


JenMysoc:
By June we would like to switch SYE onto PombolaLite instead of Pombola to give Nigeria more control over the homepage and the content of the site. 

Linked to #1812

JenMysoc:
Ok so they actually need this by the start of July

JenMysoc:
Due to go live soon now in github.com/theyworkforyou/shineyoureye/issues/34

PR #1998, opened 2016-02-23
Remove PU search logic from Nigeria views

chrismytton:
As part of the migration to running ShineYourEye as a static site (https://github.com/theyworkforyou/shineyoureye) we need to extract the PU search, which is a dynamic component, into a separate API.

This change is the incremental step that switches the existing Pombola site to use the new API.

screen shot 2016-02-23 at 19 31 15

Part of https://github.com/mysociety/pombola/issues/1914

### DIFF
--- a/pombola/nigeria/views.py
+++ b/pombola/nigeria/views.py
@@ -1,5 +1,7 @@
 import re
 
+import requests
+
 from mapit.models import Area
 
 from pombola.core.models import Place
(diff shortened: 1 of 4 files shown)


pombola/nigeria/views.py
```diff
@@ -164,29 +72,24 @@ def get_context_data(self, **kwargs):
...
 
     def parse_params(self):
         super(NGSearchView, self).parse_params()
-        tidied_as_if_pun = tidy_up_pun(self.query)
-        # Check if this is a known form of PUN:
-        if pun_re.search(tidied_as_if_pun):
-            self.pun = tidied_as_if_pun
-        else:
-            self.pun = None
+        self.pun = self.query
```
chrismytton: This logic needs to be changed slightly so we don't always assume we're dealing with a PU number.

pombola/nigeria/views.py
```diff
@@ -252,14 +155,7 @@ def get_area_from_pun(self, pun):
...
 
-        while pun:
-            try:
-                area = Area.objects.get(codes__code=pun)
-                return area
-            except Area.DoesNotExist:
-                # strip off last component
-                pun = re.sub(r'[^:]+$', '', pun)
-                pun = re.sub(r':$', '', pun)
+        return requests.get('https://pu-lookup.herokuapp.com/lookup/{}'.format(pun)).json()
```
tmtmtmtm: presumably this URL going to move out to a config entry somewhere?

chrismytton: Yep, definitely. This first commit is very much in the sling some code around to see if it works stage :)

chrismytton:
All wired up and working now:

### Before

screen shot 2016-02-26 at 17 22 46

### After

screen shot 2016-02-26 at 17 23 15

Still need to fix the tests, which are currently failing because where previously it would say no results for poll unit number for an unrecognised number it now just falls back to a regular search. I should also mock the requests that are being done in the tests so that there aren't http requests happening in the test suite.

duncanparkes:
Looks good to me, though probably worth a bit of squashing commits together to make it look simpler.

chrismytton:
[Shine Your Eye](https://www.shineyoureye.org/) is no longer running off the pombola codebase. It is now living in the [shineyoureye-sinatra repo](https://github.com/theyworkforyou/shineyoureye-sinatra). So closing this pull request as it's no longer needed.

PR #2012, opened 2016-02-24
1649 accents in search

### LINKED ISSUE #1649: searching for someone's name without its accents should still find the person
The example that PMG pointed out is that [searching for "desiree"](http://www.pa.org.za/search/?q=desiree&section=persons&order=relevance) doesn't bring up [Ms Désirée Van Der Walt](http://www.pa.org.za/person/d-van-der-walt/) when ideally her person page should be found when searching for either "desiree" or "désirée" - only [the latter](http://www.pa.org.za/search/?q=d%C3%A9sir%C3%A9e&section=persons&order=relevance) works at the moment.

It looks as if a plausible solution is to add the [`asciifolding` analyzer](http://www.elastic.co/guide/en/elasticsearch/reference/0.90/analysis-asciifolding-tokenfilter.html) to Pombola ZA's Elasticsearch index. (Those docs are for version 0.90, which we're using - later versions [mention a `preserve_original` option](http://www.elastic.co/guide/en/elasticsearch/reference/current/analysis-asciifolding-tokenfilter.html) which sounds as if it might be needed for to get searches on the accented version to still work (?))

duncanparkes:
Fixes #1649.

NOTE - needs a rebuild of the search index with 

```python manage.py rebuild_index```

### DIFF
--- a/pombola/settings/south_africa.py
+++ b/pombola/settings/south_africa.py
@@ -4,11 +4,13 @@ import os
 from .base import *  # noqa
 from .south_africa_base import *  # noqa
 
-
 HAYSTACK_CONNECTIONS['default']['EXCLUDED_INDEXES'] = [
     'pombola.search.search_indexes.PlaceIndex',
     'speeches.search_indexes.SpeechIndex',
 ]
+
+HAYSTACK_CONNECTIONS['default']['ENGINE'] = 'pombola.south_africa.search.ZAElasticSearchEngine'
+
 
 INSTALLED_APPS = insert_after(INSTALLED_APPS,
                               'markitup',
--- a/pombola/south_africa/search.py
+++ b/pombola/south_africa/search.py
@@ -10,6 +10,15 @@ class ZAElasticBackend(ElasticsearchSearchBackend):
             connection_alias, **connection_options)
 
         analyzers = self.DEFAULT_SETTINGS['settings']['analysis']['analyzer']
+
+        for analyzer_name in analyzers:
+            filter_ = analyzers[analyzer_name].get('filter')
+
+            if filter_:
+                analyzers[analyzer_name]['filter'].append('asciifolding')
+            else:
+                analyzers[analyzer_name]['filter'] = ['asciifolding']
+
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

**Comment by dracos** [comment]
As you've added asciifolding to all the existing analyzers, including snowball, do you need to create your own folding analyzer, or won't the default use of snowball then have asciifolding in it anyway?

File: pombola/settings/south_africa.py [hunk 1]
```diff
@@ -4,12 +4,17 @@
...
 from .south_africa_base import *  # noqa
 
+from pombola.south_africa.search import ZAElasticBackend
 
 HAYSTACK_CONNECTIONS['default']['EXCLUDED_INDEXES'] = [
     'pombola.search.search_indexes.PlaceIndex',
     'speeches.search_indexes.SpeechIndex',
 ]
 
+# Change to custom backend for
```
tmtmtmtm: this looks like it's missing something important…

duncanparkes: Just half a comment I think. I probably intended to remove all of it. Well spotted though!

pombola/settings/south_africa.py
```diff
@@ -4,12 +4,17 @@
 from .base import *  # noqa
 from .south_africa_base import *  # noqa
 
+from pombola.south_africa.search import ZAElasticBackend
```
duncanparkes: I wonder if I actually needed this import. Doesn't look likely.

wfdd:
It doesn't appear as though this issue is specific to South Africa, and it might be better addressed by using fuzzy queries, which would also work with transliterated names (e.g. German ö -> oe).

dracos:
@wfdd If you had German names, you would probably want to do something like make the snowball analyzer use german or german2, or use the german_normalization filter, etc. I think each instance needs to decide what is best for it.

wfdd:
Fair enough.

mhl:
This is merged now as 3eaad19b75736aab706185aab487e6b780030e3c and e79089ade244aaa271271e0087f50f232788b557

PR #2014, opened 2016-02-24
Upgrade to Django 1.7.11

mhl:
These changes upgrade Pombola to use Django 1.7.11. The corresponding
pull requests for its za-hansard and popolo-name-resolver dependencies are here:

* https://github.com/mysociety/za-hansard/pull/72
* https://github.com/mysociety/popolo-name-resolver/pull/1

When actuallly running this migration there's an error because the 0002 migration
from easy_thumbnails tries to create a table that's already been created - since
it's not the initial migration (I think) it's not faked automatically, whereas 0001 is.

When this fails, you can run:

    ./manage.py migrate --fake easy_thumbnails 0002

I found when running this locally that the sequence for django content type IDs
was out of sync, so I also needed to run:

    SELECT setval('django_content_type_id_seq', (SELECT MAX(id) FROM django_content_type));

... for the post-migration content-type creation to succeed. (I'll look out for whether
this is the case when running it on the staging server.)

There is still one obvious deprecation warning after this:

```
/home/mark/.virtualenvs/pombola-1.7/local/lib/python2.7/site-packages/djqmethod.py:44: RemovedInDjango18Warning: `Manager.get_query_set` method should be renamed `get_queryset`.
  class Manager(models.Manager, QMethodLookupMixin):
```

This is from SayIt'  use of `django-qmethod`'s QMethodLookupMixin. @dracos has done
a fix for this in the [SayIt upgrade to Django 1.8](https://github.com/mysociety/sayit/pull/504)
but we can't go to that version of SayIt until we're on Django 1.8. So, we should live with
the deprecation warning and do the upgrade to Django 1.8 soon.


duncanparkes:
:+1:

PR #2018, opened 2016-02-25
Add a basic API for the hansard app, using Django REST framework

mhl:
This adds endpoints for:

* venues
* sources
* sittings
* entries

If you get an individual sitting, it inlines all the entries for that sitting.  When
we've upgraded to Django 1.8, we can also add filtering of the results
with this branch: https://github.com/mysociety/pombola/tree/1817-simple-hansard-api-with-filtering

This PR should be merged after the Django 1.7 upgrade has been
deployed.

api-sitting-instance


pombola/hansard/api/views.py
```diff
@@ -0,0 +1,59 @@
...
+
+
+class EntryViewSet(viewsets.ModelViewSet):
+    queryset = Entry.objects.order_by('id')
+    serializer_class = EntrySerializer
+
+
+class SittingViewSet(viewsets.GenericViewSet):
+
+    queryset = Entry.objects.order_by('id')
```
struan: Should this be here?

struan:
Other than the queryset on the sittings view this looks fine to me code wise.

pombola/hansard/api/views.py
mhl: If you don't have it, then going to /api/v0.1/hansard/sittings/ gives you this error:

```
'SittingViewSet' should either include a `queryset` attribute, or override the `get_queryset()` method.
```

The traceback is:

```
Traceback:
File "/home/mark/.virtualenvs/pombola-1.7/local/lib/python2.7/site-packages/django/core/handlers/base.py" in get_response
  137.                 response = response.render()
File "/home/mark/.virtualenvs/pombola-1.7/local/lib/python2.7/site-packages/django/template/response.py" in render
  103.             self.content = self.rendered_content
File "/home/mark/.virtualenvs/pombola-1.7/local/lib/python2.7/site-packages/rest_framework/response.py" in rendered_content
  71.         ret = renderer.render(self.data, media_type, context)
File "/home/mark/.virtualenvs/pombola-1.7/local/lib/python2.7/site-packages/rest_framework/renderers.py" in render
  675.         context = self.get_context(data, accepted_media_type, renderer_context)
File "/home/mark/.virtualenvs/pombola-1.7/local/lib/python2.7/site-packages/rest_framework/renderers.py" in get_context
  654.             'filter_form': self.get_filter_form(data, view, request),
File "/home/mark/.virtualenvs/pombola-1.7/local/lib/python2.7/site-packages/rest_framework/renderers.py" in get_filter_form
  592.         queryset = view.get_queryset()
File "/home/mark/.virtualenvs/pombola-1.7/local/lib/python2.7/site-packages/rest_framework/generics.py" in get_queryset
  67.             % self.__class__.__name__

Exception Type: AssertionError at /api/v0.1/hansard/sittings/
Exception Value: 'SittingViewSet' should either include a `queryset` attribute, or override the `get_queryset()` method.
```

struan: Sorry, I should have been more specific. It was the fact that it was in the `SittingViewSet` but the queryset is for `Entry` which was confusing about it.

struan:
Looks good to me.

pombola/hansard/api/views.py
mhl: It's used in constructing the the filter form, but we're not using filters at the moment. I think we can ignore this for the moment - I've added a comment explaining that in a fixup commit.

mhl:
Merged and deployed now

PR #2039, opened 2016-03-01
Remove core_import_popolo and update related docs

### LINKED ISSUE #698: core_import_popolo dies on different PositionTitle
If a position_title_name has the same slug as one already in the system (e.g. there's been a change in role name that is only in punctuation/spacing), this script throws an error.

### LINKED ISSUE #1078: core_import_popolo script does not update details
Noticed when re-running script on dev vhost that the full party names added in #966  were not being applied. The [core_import_popolo.py](https://github.com/mysociety/pombola/blob/master/pombola/core/management/commands/core_import_popolo.py) script does have a `--delete-old` argument, but I'm not sure what effect this would have on the related data such as hansard etc.

I think that the expected behaviour is that if `--commit` is given names etc should be up dated for records found in the database to match what is in the popolo file.

### LINKED ISSUE #2031: Remove core_import_popolo
Related to #2029  and #2030; also the reference on the API help page should be updated.

chrismytton:
We no longer need this command as the recommended way of creating a
development environment is now to use database dumps.

Also removes references to it in various bits of documentation and code
comments.

Fixes #698 
Fixes #1078 
Fixes #2031


tmtmtmtm:
I'm assuming this also closes https://github.com/mysociety/pombola/issues/698 and https://github.com/mysociety/pombola/issues/1078?

chrismytton:
@tmtmtmtm Yes, looks like it does. I've updated the issue text to reflect this.

mhl:
:+1:

Issue #2019, opened 2016-02-25
KE: Cabinet positions


tmtmtmtm:
Currently, on their pages for cabinet ministers we don't show much info:
![screen shot 2016-02-25 at 14 51 44](https://cloud.githubusercontent.com/assets/57483/13323089/c54f1f20-dbcf-11e5-8960-6ec137a52132.png)

until you click through to full experience:

![screen shot 2016-02-25 at 14 52 00](https://cloud.githubusercontent.com/assets/57483/13323091/c7155496-dbcf-11e5-8cf1-af6120704fa7.png)

this seems to be because of a better separate of powers, where cabinet members aren't in the Assembly, and the simplest solution might just be to also pick out Cabinet positions to highlight on the main page, like we do with Assembly positions.

tmtmtmtm:
Note that a similar thing happens for people with only historic positions: https://github.com/mysociety/pombola/issues/1983

PR #2057, opened 2016-03-02
[KE]: Fix misleading "No experience" claim

### LINKED ISSUE #1983: KE: Misleading "No experience" claim
On, for example http://info.mzalendo.com/person/adelina-mwau/, we say under Experience "No records found.":

screen shot 2016-02-23 at 13 47 26

This is quite odd, as we do actually know quite a lot about their Experience:

screen shot 2016-02-23 at 13 48 06
 

If the issue is that this section is only for *current* information, then we should say so explicitly (e.g. "No current positions known"), though it would be even better if we could include the most important historic position(s) here.

chrismytton:
We were saying people had no experience when actually what we meant was
the person had no current positions that we know about. This changes the
text on person pages to make that clearer.

## Person page

screen shot 2016-03-02 at 14 18 01

## Person's experience page

screen shot 2016-03-02 at 14 18 08

Fixes #1983 (paired with @duncanparkes)

### DIFF
--- a/pombola/core/templates/core/person_position_section.html
+++ b/pombola/core/templates/core/person_position_section.html
@@ -38,6 +38,6 @@
         {% endif %}
     </li>
 {% empty %}
-    <li>No records found.</li>
+    <li>No current positions known.</li>
 {% endfor %}
 </ul>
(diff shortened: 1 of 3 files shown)


tmtmtmtm:
Working out what we _should_ say when there are no memberships (we should probably still highlight _something_ here) can be deferred to https://github.com/mysociety/pombola/issues/2019

tmtmtmtm:
:+1:

PR #2074, opened 2016-03-03
2020 organisation merge command

### LINKED ISSUE #2020: add a core_merge_organisations command
We don't need to merge organisations as often as we need to merge people, but it does come up - see #1974 and #1940 - it would be good to add a command that does that, similarly to core_merge_people and creates slug redirects.

duncanparkes:
Nearly ready - just needs slugredirect checking on the views.

Closes #2020


chrismytton:
Just tried merging two organizations locally and it worked fine, so this gets a :+1: from me once @duncanparkes is happy with the slug redirect bits.

Issue #2048, opened 2016-03-02
Add Open Graph tags to place pages


chrismytton:
Just like https://github.com/mysociety/pombola/issues/1947

Issue #2047, opened 2016-03-02
Add Open Graph tags to organization pages


chrismytton:
Just like https://github.com/mysociety/pombola/issues/1947

Issue #1956, opened 2016-02-15
add Open Graph meta data for homepage / "any other" page


mhl:
### Front page (or any other generic page on the site):

*  a "keep track of your politicians" comment - will depend a lot on the attitude to politicians int he contry

Issue #1955, opened 2016-02-15
add Open Graph meta data for Info pages


mhl:
### Info page:

* Add a box in the edit interface so you can add a snippet that will be shared

PR #2076, opened 2016-03-03
Open Graph tags

chrismytton:
This adds some default Open Graph tags to all pages on the website. It also adds specific Open Graph tags for Organizations, Places and blog posts.

Fixes https://github.com/mysociety/pombola/issues/2047
Fixes https://github.com/mysociety/pombola/issues/2048
Fixes https://github.com/mysociety/pombola/issues/1956
Fixes https://github.com/mysociety/pombola/issues/1955

### DIFF
--- a/pombola/core/context_processors.py
+++ b/pombola/core/context_processors.py
@@ -1,4 +1,5 @@
 from django.conf import settings
+from django.contrib.sites.models import Site
 
 
 def add_settings( request ):
@@ -23,3 +24,7 @@ def add_settings( request ):
             'FACEBOOK_APP_ID':              settings.FACEBOOK_APP_ID,
         }
     }
+
+
+def site_processor(request):
+    return {'site': Site.objects.get_current()}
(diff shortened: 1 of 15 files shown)


mhl:
I think using `og:type` of `profile` is right for `Person` but not `Organisation` and `Place`. There is a `place` type, but I'm not sure what's right for `Organisation`.

mhl:
I've left a few comments which would be worth addressing, but this looks good.

Could you also squash this down into fewer commits? (Some are essentially fixups to earlier commits - e.g. perhaps just one for everything to do with people, one for places, etc.)

chrismytton:
@mhl I've fixed the things in your comments so I think this is ready to go now.

mhl:
@chrismytton Sorry to say this at this stage, but I think these could really easily be broken without us noticing in the future - could you add some tests to check that the expected tags are returned for each page?

chrismytton:
Tests added, commits rebased. Once Travis is happy I'll get this merged/deployed.

chrismytton:
Oh dear, accidentally merged with a WIP commit which I had to sneak past the update hook on git.mysociety.org, sorry! :disappointed:

chrismytton:
This is how it looks on the live sites:

screen shot 2016-03-11 at 12 46 17

Issue #1866, opened 2015-12-03
sort_name should not start with a leading space.


duncanparkes:
A `sort_name` with a leading space causes someone to be put at the start of lists of people, out of order.

We should strip `sort_name` when people are added or changed. We should also go over the existing databases and strip current `sort_name`s.

duncanparkes:
(an example of a bug caused by this is in #1865)

tmtmtmtm:
There are lots of other fields where this is true and problematic also (e.g. People and Organisation names https://github.com/mysociety/pombola/issues/1972) — this should be a more general thing than this.

chrismytton:
Looks like `sort_name` [is already getting stripped when it's set in the signal handler](https://github.com/mysociety/pombola/blob/ec6b5bd9bea2ad7e133b2f75d8726f334f30517d/pombola/core/models.py#L512). Tracing through history reveals that this was added in https://github.com/mysociety/pombola/commit/487606ed6e8e684362756d26b64292dcded0e2a8 so it's possible that the bad data was introduced before then?

Leaving this ticket open as the old data still needs to be cleaned up.

mhl:
@chrismytton That's only responding to the `post_init` signal, so any edits to `sort_name` might introduce spaces.

chrismytton:
I've tidied up the whitespace on all Nigerian models as part of https://github.com/mysociety/pombola/issues/1972. Still need to run the commands from https://github.com/mysociety/pombola/issues/1972#issuecomment-193322961 against KE and ZA.

PR #2077, opened 2016-03-03
Models can define fields to be whitespace stripped

chrismytton:
This adds a 'fields_to_whitespace_normalize' property to models which
can be set to a list of fields which should have their whitespace
removed before the model is saved.

First part of #1866

### DIFF
--- a/pombola/core/models.py
+++ b/pombola/core/models.py
@@ -53,6 +53,18 @@ class ModelBase(models.Model):
     created = models.DateTimeField( auto_now_add=True )
     updated = models.DateTimeField( auto_now=True )
 
+    fields_to_whitespace_normalize = []
+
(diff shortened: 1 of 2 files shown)


mhl:
Looks good so far, I guess you're going to add lots of other fields to be normalized as well.

It'd be good to also do `re.sub(r'(?ms)\s+', ' ')` as part of normalizing the whitespace too...

chrismytton:
> It'd be good to also do re.sub(r'(?ms)\s+', ' ') as part of normalizing the whitespace too...

Ooh, yes good idea!

chrismytton:
@mhl I've added the regex you suggested and defined some more fields to have their whitespace normalized, so I think this is ready to go now!

mhl:
LGTM :+1:

PR #2079, opened 2016-03-03
Get rid of 'patterns' function from urls files.

duncanparkes:
Calls to the patterns function can be replaced with just lists
from Django 1.8.

https://docs.djangoproject.com/es/1.9/releases/1.8/#django-conf-urls-patterns


mhl:
:+1: Cherry-picked to master as 8ef9d29963e809e62e2e89fa628eaa - thanks!

PR #2082, opened 2016-03-04
WIP: add a pop-up 'add an organisation' from inline position admin

mhl:
I'm not sure we actually want to do this, but the WIP pull request
is just so it doesn't get lost...


chrismytton:
There hasn't been much activity on this pull request for a couple of years, so I'm going to close it. Please do feel free to reopen if you think this is a useful feature!

Issue #1905, opened 2016-01-14
Installs to link back to mySociety


MyfanwyNixon:
A point from the International Copy discussion day: https://docs.google.com/document/d/1MKHqc_fywwtRjLmiQpaIlTG-0JhnV82EcgyM1QOyst0/edit

>All Pombola installs should link back to a place where people can find out more about the software that powers them.

It looks as if this place will be a revamped page on the mySociety website in the near future; watch this space.

I assume the traditional placement for such a link would be in the footer, along the lines of "Powered by Pombola, from mySociety".

tmtmtmtm:
For Pombola the issue is primarily that the current footer links to the Github repo, which isn't a very good introduction for people, and should be replaced by a link to a Pombola-specific page (once that exists, and explains the concept within the wider stream, etc.)

mhl:
It could link to this page instead: https://www.mysociety.org/projects/parliamentarymonitoring/pombola/

MyfanwyNixon:
Yep, that seems right.

PR #2086, opened 2016-03-04
Link to a more descriptive page in the footer

chrismytton:
Rather than linking to the GitHub repo link to the mySociety site which
has a better explanation of Pombola and the problems it solves.

Fixes https://github.com/mysociety/pombola/issues/1905


duncanparkes:
:+1:

Issue #1902, opened 2016-01-13
remove old kuvakazim vhosts


mhl:
If the "pombola light" static Kuvakazim site is working out well, we should remove the old Pombola-based Kuvakazim vhosts - hopefully @tmtmtmtm or @chrismytton can say when they're happy for that to happen.

tmtmtmtm:
I suspect we're OK to go ahead with this now? @chrismytton ?

chrismytton:
Yep, fine by me. I believe I've copied all the relevant files over to the [`media_root` directory in KuvakaZim](https://github.com/mysociety/kuvakazim/tree/master/media_root), let me know if I've missed anything!

chrismytton:
Done. Vhost files and database dumps are in my home directory on firefly.

PR #2087, opened 2016-03-04
Remove all code related to Zimbabwe

chrismytton:
Following on from #1902


duncanparkes:
:+1: Looks good to me. We should make sure we remove the search index.

PR #2097, opened 2016-03-09
Send update_hansard cron emails to mzalendo-managers

### LINKED ISSUE #1582: KE: make the output of update_hansard go to a different email alias
These haven't been going to Mzalendo admins for a while, since the cron-pombola@mysociety.org alias was getting a lot of error emails at one point, and they asked to be taken off the alias. Unfortunately, the update_hansard output (with lists of unmatched names) was still going to cron-pombola@mysociety.org. We should have a separate email alias that gets the emails that can be actioned by non-developers...

chrismytton:
This alias is logged in Google Groups and goes to Jess and the rest of
the Mzalendo team.

Fixes #1582


duncanparkes:
:+1:

PR #2118, opened 2016-03-21
Improvements to core_find_stale_elasticsearch_documents

### LINKED ISSUE #2084: add to core_find_stale_elasticsearch_documents an option to remove stale ES documents
At the moment, the `core_find_stale_elasticsearch_documents` script will find documents that are in Elasticsearch for which the corresponding Django object has been deleted (but somehow a signal wasn't sent so that Haystack updated ES - there's an issue about the underlying problem: https://github.com/mysociety/pombola/issues/1424)

If we want to fix this at the moment you need to run `./manage.py rebuild_index`, which is time-consuming and (because we haven't implemented [zero-search-downtime reindexing of Elasticsearch yet](https://github.com/mysociety/pombola/issues/1584#issuecomment-72438116)) means search won't work properly in the mean time.

It would be convenient if there were a `--fix` option to `core_find_stale_elasticsearch_documents` which would remove any of the stale documents it finds in Elasticsearch, so we could simply fix this issue when it comes up.

(This is annoying because the autocomplete in the search box breaks and sends error emails when one of these cases comes up. That could also be worked around, but for the moment we'd rather know when the data's got out of sync so we can investigate further how that's happened.)

mhl:
The first two commits improve `core_find_stale_elasticsearch_documents` as a
tool for investigating the stale search index entries problem. The final commit
adds a option for fixing those issues as they're found. More background from
the commit message below:

There is an as yet unsolved problem where an object in the database
has been deleted, but there is still a stale Elasticsearch document
that refers to it; for some reason Haystack hasn't got the signal
to remove it.

We were fixing this synchronisation problem by running
`./manage.py rebuild_index`, but that takes a really long time to run
and means search will be missing results while it's running (since it
clears all the indices, and we haven't implemented
zero-search-downtime-reindexing of Elasticsearch yet).  This commit
adds a `--delete` option to `core_find_stale_elasticsearch_documents` so
that it can also remove any stale documents, which will mean we can
fix the symptoms of the problem without search downtime.

n.b. in looking through Haystack's source in the course of fixing
this issue, it looks from the source code as if
`./update_index --remove <MODEL>` would be a much quicker way of
fixing stale objects for a particular model - this commit essentially
duplicates that functionality.  However, this command is still useful
as a way of investigating the stale documents, and the tiny amount of
extra code here to let it fix that problem as well seems worth adding
to me.

Fixes #2084

### DIFF
--- a/pombola/core/management/commands/core_find_stale_elasticsearch_documents.py
+++ b/pombola/core/management/commands/core_find_stale_elasticsearch_documents.py
@@ -17,18 +17,22 @@ def get_all_indexed_models():
     available_models = {}
 
     for backend_key in backends:
+        connection = haystack_connections[backend_key]
+        backend = connection.get_backend()
         unified_index = haystack_connections[backend_key].get_unified_index()
         for app in haystack_load_apps():
             for model in haystack_get_models(app):
                 try:
-                    unified_index.get_index(model)
+                    index = unified_index.get_index(model)
                 except NotHandled:
                     continue
                 model_name = model.__module__ + '.' + model.__name__
                 available_models[model_name] = {
                     'backend_key': backend_key,
+                    'backend': backend,
                     'app': app,
                     'model': model,
+                    'index': index,
                 }
 
     return available_models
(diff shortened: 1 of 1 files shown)


duncanparkes:
:+1:

PR #2121, opened 2016-03-24
Switch to using django-images from PyPI.

### LINKED ISSUE #1898: migrate to using mysociety-django-images
We extracted pombola's `images` app into https://github.com/mysociety/django-images (mysociety-django-images on pypi) for YNR, however we didn't switch Pombola to use that at the same time. We should do that - it should be a quick thing to do since nothing's changed in the models.

duncanparkes:
I've brought the small number of changes in pombola/images/ since django-images was spun off into django-images version 0.0.6, so we can now switch to using that.

Fixes #1898


duncanparkes:
Rebased and merged - see 8739e4f2ea9a0aebe5ece7ddb052ab2abbe27827.

PR #2125, opened 2016-04-01
Greatly reduce queries needed for Place pages in the admin

### LINKED ISSUE #2085: admin page for Place does thousands of queries
The drop-down boxes for organisations and [one of the other drop-downs] include objects whose `__unicode__` methods include the `OrganisationKind` and the `ParliamentarySession`, and an extra query is made for each of those (about 6000 queries when I just tried).

The queryset that's used by the admin interface should be changed to `select_related` on these related fields.

For bonus points, check whether there's a similar issue for any other models in the admin, using Django Debug Toolbar to inspect the queries being made.

mhl:
Place pages (both the list and the detail) in the admin were incredibly
slow because they were making thousands of queries. This is because the
__unicode__ methods of Place and Organisation use related objects.

There are two changes here:

 * get_queryset - This is used for the list page.

 * get_field_queryset - This is used to create the querysets for
   populating the select elements for foreign keys on the detail
   page.

So the former speeds up the place page, while the latter speeds up the
detail page.

Fixes #2085

### DIFF
--- a/pombola/core/admin.py
+++ b/pombola/core/admin.py
@@ -211,6 +211,19 @@ class PlaceAdmin(StricterSlugFieldMixin, admin.ModelAdmin):
         IdentifierInlineAdmin,
         )
 
+    def get_field_queryset(self, db, db_field, request):
+        if db_field.name == 'parent_place':
+            fields = ('kind', 'organisation', 'parliamentary_session')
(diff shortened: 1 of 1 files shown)


chrismytton:
Looks good! :rocket: :+1:

PR #2126, opened 2016-04-08
Make Pombola deployable on Heroku

mhl:
This isn't quite working yet - the idea was to make Pombola deployable
on Heroku so we can use its nice Review Apps feature.

At the moment it fails, though - the bundle appears to be created but
when django-pipeline invokes compass, we get the error:

```
pipeline.exceptions.CompilerError: /tmp/build_6c0606c358819e8b9bd8ed59ff32e233/vendor/ruby-2.0.0/lib/ruby/2.0.0/rubygems/core_ext/kernel_require.rb:55:in `require': cannot load such file -- bundler/setup (LoadError)
```

Full log here: https://gist.github.com/mhl/94589d97ba58e3e09c26f32607f9c78f


chrismytton:
👍  All works for me! Looks like Travis is failing because it's using Ruby 1.9.3 (!) but the Gemfile now specifies 2.0.0, not sure if there's a way to work around that?

mhl:
@chrismytton I've removed the Ruby version specification from the Gemfile (it was only there in the hope that that would fix the bundler issue) and that seems to be fine with both Heroku and Travis.

chrismytton:
:+1: LGTM!

PR #2127, opened 2016-04-08
Switch to using the PyPi-packaged django-slug-helpers

mhl:


duncanparkes:
Personally, I would move all the import lines for slug helpers to above the pombola ones to reflect that it's now an external packages, but I don't feel overly strongly about this.

:+1:

duncanparkes:
Rebased and merged - see a56f4f81e0a48fe95ed86e7068e8c5fd87081999

PR #2140, opened 2016-04-20
Make sure JS files from addExtraJS are loaded after in-page <script>s

### LINKED ISSUE #2137: [ZA] Ward Rep locator tabs not working correctly
On the new Ward Rep Locator, when you type in "Cape Town" the search is returning all the results for all the types of representative (Ward Councillor, MP and MPL) on one screen. 

When you click the tab to see a different kind of representative, for example MP instead of Ward Councillor, you are simply moved down the page to the point in the list of representatives where that type of representative starts to appear.

screen shot 2016-04-20 at 10 09 07

What should happen is when you click on the Ward Councillor tab it should ONLY display Ward Councillors, the MP tab should only display MPs. etc

### LINKED ISSUE #2138: [ZA] Map not showing on Ward Rep Locator results page
When you search for Ward Councillors using the Rep Locator the results are being returned without the map which usually accompanies them. Instead there is a large white space.

screen shot 2016-04-20 at 10 08 35

### LINKED ISSUE #2139: [ZA] Information text about the different types of representative not showing on Rep locator results
When searching for representatives in an area using the Rep Locator you expect to see a small piece of informational text which explains the reason you might want to contact each different type of representative. This text was showing above the map and just under the tabs for the different representative types.

This text is currently not showing when the rep locator returns results.

screen shot 2016-04-20 at 10 08 35

mhl:
In production we were getting an error that suggested that tabs.js was
being loaded before map.js.

These are loaded by different mechanisms (warning: this is all awful):

 - map.js is loaded via `{% javascript 'google-map' %}` in the `<body>`
   which is replaced by a `<script>` tag that loads it directly

 - tabs.js is loaded from the 'both' case of yepnope (although in a
   slightly convoluted way, because of the hack to make django-pipeline
   work with yepnope) - it starts with `addExtraJS`

I believe this is happening because the files that are loaded from
'both' don't wait for the document to be ready, so the `<script>` tag in
the body might not have been loaded yet.

This moves the loading of the extra JS into the `complete` callback of
yepnope, and wraps the body of that function in `$(function() { })` so
that it really is only run when the document's ready.

n.b. This is just a quick fix because the rep locator page is broken on
People's Assembly. I think the best way to fix this ultimately is to
remove this tangled web of hacked-together delayed loading entirely,
and replace it with something more coherent.

Fixes #2137
Fixes #2138
Fixes #2139

### DIFF
--- a/pombola/core/static/js/loader.js
+++ b/pombola/core/static/js/loader.js
@@ -24,14 +24,14 @@
                 '//ajax.googleapis.com/ajax/libs/jquery/1.7.1/jquery.min.js',
             ].concat(pipeline_desktop_only),
             both: pipeline_desktop_and_mobile.concat(
-                ['//www.google.com/jsapi']).concat(extra_js),
+                ['//www.google.com/jsapi']),
             complete: function () {
-                for (i=0; i<pombola_run_when_document_ready_array.length; i++) {
-                    $( pombola_run_when_document_ready_array[i] );
-                }
-
-                // Now load all the optional bits that we didn't want slowing down the more important bits
-                Modernizr.load(pipeline_analytics);
+                $(function() {
+                    for (i=0; i<pombola_run_when_document_ready_array.length; i++) {
+                        $( pombola_run_when_document_ready_array[i] );
+                    }
+                    Modernizr.load(extra_js.concat(pipeline_analytics));
+                });
             }
         }
     );


duncanparkes:
Looks like it works OK on staging, and the javascript change looks fine to the extent I understand it :+1:

PR #2147, opened 2016-04-26
To help the EP scraper, add the person ID in a <meta> tag

### LINKED ISSUE #2130: expose Pombola person IDs somewhere within person pages
This is so that EveryPolitician can store these as an additional identifier column,
so that we can easily match people once we're getting membership information
from EveryPolitician.

mhl:
We'd like the EveryPolitician scraper to be able to extract the Pombola
person ID in a predictable way - this commit adds that value in a
`<meta name="pombola-person-id">` in the `<head>`.

Fixes #2130


chrismytton:
Looks good! 👍

PR #2150, opened 2016-04-28
Output more contact details

### LINKED ISSUE #2136: More contact details on the Rep locator results page.
Megan suggests showing the following extra data where applicable: Name of Ward; Municipality name, website, and phone; Party name, website, and phone.

Relates to #1895.

duncanparkes:
Some more contact details for @zarino to use.
- Ward number
- Municipality name, website, telephone
- Party name, website, telephone

There are more Municipality and Party contact details which could
be added if that's the way people want to go...

Closes #2136.


zarino:
It now looks like this:

screen shot 2016-05-04 at 16 33 21

Feels quite busy, but there's only ever one councillor visible at a time, so I think we can get away with it.

mhl:
One small thing is that sometimes there are multiple phone numbers separated by `; `

ward-councillor-more-details

... which should be split into multiple phone numbers, I think.

However, let's deploy this anyway, and create an issue for that.

zarino:
@mhl: Yes, split those into multiple `<a>` elements, inside the same `<li>`, I think.

zarino:
Or they could be separate `<li>`s, meaning you'd have one number per line, and a little handful of phone icons stacked on top of one another. Which might be more obvious?

(It'll still be a bit weird, because there'll be no indication of which number you're meant to call – but I guess that's a gap in the source data and we can't easily fill it?)

mhl:
I've created an issue for that here: https://github.com/mysociety/pombola/issues/2153 and copied over @zarino's comments

PR #2163, opened 2016-05-31
Track clicks on jQuery tabs in Google Analytics

mhl:
We don't have any data about how many people try changing tab on the
person pages or rep locator pages; this event tracking code should
record that data.

### DIFF
--- a/pombola/south_africa/static/js/tabs.js
+++ b/pombola/south_africa/static/js/tabs.js
@@ -34,6 +34,25 @@ $(function() {
     active: activeTabIndex
   });
   $(".tabs").on("tabsactivate", function(event, ui) {
-    history.pushState(null, null, '#' + ui.newPanel[0].id);
+    var panelID = ui.newPanel[0].id, gaCategoryPrefix;
+    history.pushState(null, null, '#' + panelID);
+    // If Google Analytics is loaded, try to track clicks on the tabs
+    if (typeof ga === 'function') {
+      // We only use jQuery tabs on the rep locator page and person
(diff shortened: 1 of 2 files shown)


pombola/south_africa/static/js/tabs.js
```diff
@@ -34,6 +34,23 @@ $(function() {
...
+      // cases with the event category:
+      if ($('.rep-locator-tabs').length) {
+        gaCategoryPrefix = 'rep-locator-tab-';
+      } else {
+        gaCategoryPrefix = 'tab-';
+      }
+      ga('send', {
+        hitType: 'event',
+        eventCategory: gaCategoryPrefix + panelID,
+        eventAction: 'activate',
```
zarino: Trailing comma here will throw a syntax error on IE7 and below.

pombola/south_africa/static/js/tabs.js
```diff
@@ -34,6 +34,23 @@ $(function() {
...
+    var panelID = ui.newPanel[0].id, gaCategoryPrefix;
+    history.pushState(null, null, '#' + panelID);
+    // If Google Analytics is loaded, try to track clicks on the tabs
+    if (typeof ga === 'function') {
+      // We only use jQuery tabs on the rep locator page and person
+      // pages for South Africa (at the moment); distinguish those
+      // cases with the event category:
+      if ($('.rep-locator-tabs').length) {
+        gaCategoryPrefix = 'rep-locator-tab-';
+      } else {
```
zarino: Why not have a `person-page-` prefix here for the person pages?

PR #2168, opened 2016-06-02
Simplify Pombola's JavaScript loading

mhl:


chrismytton:
Woo, great to see the Pombola JavaScript finally simplified like this. LGTM! 👍

Issue #2092, opened 2016-03-09
KE: Inconsistent linking of legislative memberships


tmtmtmtm:
Any "old-style" Member of Parliament has "Parliament" link through to the Organization page for the Parliament. However "new-style" "Member of the National Assembly" has no equivalent link to the National Assembly page:

![screen shot 2016-03-09 at 12 27 37](https://cloud.githubusercontent.com/assets/57483/13635243/681f9ae6-e5f2-11e5-872f-2296da2bb68b.png)

PR #2172, opened 2016-06-07
Add filters for viewing positions from particular parliamentary sessions

mhl:
This will produce slightly confusing results at the moment because the 2007 -> 2013 parliamentary session object is associated with the National Assembly organisation, when really it should be associated with Parliament with the current way it's modelled. https://github.com/mysociety/pombola/issues/2092 is related. However, we can fix that after deploying this, I think.

non-current

current

### DIFF
--- a/pombola/core/management/commands/core_export_to_popolo_json.py
+++ b/pombola/core/management/commands/core_export_to_popolo_json.py
@@ -11,8 +11,6 @@ from pombola.core.popolo import get_popolo_data
 from django.core.management.base import BaseCommand, CommandError
 
 
-ideal_collection_order = ('organizations', 'persons', 'posts', 'memberships')
-
 class Command(BaseCommand):
     args = 'OUTPUT-DIRECTORY POMBOLA-URL'
     help = 'Export all people, organisations and memberships to Popolo JSON and mongoexport format'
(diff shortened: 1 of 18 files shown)


tmtmtmtm:
@mhl it would be good if you could add some screenshots to this

mhl:
Oops, I've done that now, @tmtmtmtm - and added a comment about needing to change the earlier ParliamentarySession object.

mhl:
For example, to find the 2007-2013 members of parliament, I think that would be:

* http://mzalendo.staging.mysociety.org/position/mp/governmental/parliament/?session=na2007

Whereas the people who had positions at any time in the latest National Assembly session is:

* http://mzalendo.staging.mysociety.org/position/member-national-assembly/?session=na2013

Since Jessica is keen that we don't change the modelling of the parliament, we could override the presentation to link back to the old parliament for Kenya's National Assembly position pages.

chrismytton:
Looks good to me! 👍

tmtmtmtm:
Is that "31 December 9999" something we already have in other places, or is that an internal representation leaking out to public view?

mhl:
@tmtmtmtm Oops, yes, we should show "present" instead of that here.

tmtmtmtm:
@mhl whilst you're fixing that it might be useful to add a couple of screenshots of how this looks on ZA as well.

mhl:
@tmtmtmtm To respond to your question about South Africa screenshots, currently there are no `ParliamentarySession` objects for South Africa, so that site's appearance will be unchanged. It probably should have some, but I'd rather deal with that as a separate issue.

tmtmtmtm:
@mhl: were the ParliamentarySessions for South Africa ever added? The import of South African data into EveryPolitician is broken at the moment, and it would be good to be able to handle that in an equivalent manner to how we're working with Kenyan data.

PR #2173, opened 2016-06-14
2170 za homepage tweaks

### LINKED ISSUE #2170: update the People's Assembly homepage
As background, PMG got some advice from a social media/digital marketing company that said that:

> the mandate of the PA website should be
> clearer and more prominent on the homepage. The green band should be solely
> about connecting with your rep and that is why the message should be
> centered and the levels of representation should be included (See attached).

A summary of the changes they've asked for is:

> 1) Center the Rep Locator text on the green band and insert 'At local,
> provincial and national level' underneath 'Contact your representative'. The
> copy can be bigger across the green band.
> 2) The email sign-up should be an icon next to the Facebook icon at the top
> of the page (see rough idea on the attached screenshot). It should be
> similar in size and colour of the Facebook and twitter icons.
> 3) Remove the 'Subscribe' link at the top of the page (circled in red),
> because it links to an outdated sign-up page for the PMG newsletter.

The mock-up they sent us was:

homepage

(I think I misunderstood this request initially; in an earlier email it sounded as if there were two main calls-to-action: enter your address to find your representative, and "subscribe" to their Facebook feed, but I realise they meant the "subscribe" link should still refer to the email sign-up, and there's one main call-to-action: enter you address to find your representative.)

If there's more than one idea from the design team about how to implement this, I think this would be a great case for A/B testing, since there's a clear and easily measurable outcome we're looking for.

zarino:
Fixes #2170.

* Adds newsletter signup icon to header and footer.
* Standardises hover effect on social icons.
* Standardises social icon layout between header and footer.
* Social icons in header no longer hidden on small screens.
* "Contact your representative" section of homepage now occupies less vertical space on wide screens.

screen shot 2016-06-14 at 09 47 21

screen shot 2016-06-14 at 09 47 49

screen shot 2016-06-14 at 09 47 39

screen shot 2016-06-14 at 09 48 20


mhl:
Looks good, thank-you @zarino :+1: From the email thread, I'm still not 100% convinced that there isn't a different mailing list they're meaning us to link to, but I guess we'll know soon enough :) That's live now.

PR #2175, opened 2016-06-20
ZA: add a 3 second timeout to HTTP requests to external services

### LINKED ISSUE #2174: ZA: add short timeouts for all external API requests
These may be causing the site to go down when the external APIs are very slow to respond (e.g. due to high packet loss) - we can do this simply with requests:

* http://docs.python-requests.org/en/master/user/quickstart/#timeouts

Although there is also the question of what's a reasonable timeout. A couple of seconds, perhaps?

mhl:
The site has gone down a few times recently, and we suspect this was
because of workers being tied up waiting for external services that are
timing out.  To hopefully reduce the effect of such outages, this commit
times-out any such requests after three seconds.

Fixes #2174

### DIFF
--- a/pombola/south_africa/views.py
+++ b/pombola/south_africa/views.py
@@ -59,6 +59,10 @@ from django.core.paginator import Paginator, EmptyPage, PageNotAnInteger
 from django_date_extensions.fields import ApproximateDate
 
 
+# For requests to external APIs, timeout after 3 seconds:
+API_REQUESTS_TIMEOUT = 3.05
+
+
 logger = logging.getLogger('django.request')
 
 
@@ -162,7 +166,10 @@ class LatLonDetailBaseView(BasePlaceDetailView):
         url = 'http://mapit.code4sa.org/point/4326/{lon},{lat}?type=WD'.format(
             lon=location.x, lat=location.y
         )
-        r = requests.get(url)
+        try:
+            r = requests.get(url, timeout=API_REQUESTS_TIMEOUT)
+        except requests.exceptions.RequestException as e:
+            raise WardCouncillorAPIDown(u"MapIt request failed: {0}".format(e))
         mapit_json = r.json()
         if not mapit_json:
             return []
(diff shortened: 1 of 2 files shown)


pombola/south_africa/views.py
```diff
@@ -2115,7 +2125,7 @@ def download_attendance_data(self):
         if results is None:
             results = []
             while next_url:
-                resp = requests.get(next_url)
+                resp = requests.get(next_url, timeout=API_REQUESTS_TIMEOUT)
```
dracos: No catch on this one?

pombola/south_africa/views.py
```diff
@@ -692,7 +701,8 @@ def store_or_get_pmg_member_id(self, scheme='za.org.pmg.api/member'):
             url_fmt = "https://api.pmg.org.za/member/?filter[pa_link]={}"
             api_search_url = url_fmt.format(pa_link)
             try:
-                search_resp = requests.get(api_search_url)
+                search_resp = requests.get(
+                    api_search_url, timeout=API_REQUESTS_TIMEOUT)
             except requests.RequestException:
```
dracos: Should this (and other similar ones) also be requests.exceptions.RequestException?

pombola/south_africa/views.py
```diff
@@ -59,6 +59,10 @@
 from django_date_extensions.fields import ApproximateDate
 
 
+# For requests to external APIs, timeout after 3 seconds:
+API_REQUESTS_TIMEOUT = 3
```
dracos: http://requests.readthedocs.io/en/master/user/advanced/#timeouts suggests (if the example is anything to go by?) 3.05, but meh :)

pombola/south_africa/views.py
mhl: There wasn't one previously for other types of failure, so I was going to leave the behaviour here the same.

pombola/south_africa/views.py
mhl: Oops, yes - thanks.

pombola/south_africa/views.py
mhl: Oh yes, thanks - I've changed that.

PR #2178, opened 2016-06-22
Add a simple IEBC office lookup form

mhl:
This adds a management command for building a Python data file with data
on the IEBC offices we know about:

   kenya_parse_iebc_office_data

The result of that is committed as:

   pombola/kenya/election_data_2017/iebc_offices.py

This commit also introduces views, forms and templates for a simple office
lookup widget; a Select2 enhanced drop-down of constituencies grouped by
county.

[An alternative approach would be to create a PlaceKind representing IEBC
 offices, and then Place objects of that kind that represent the location
 of the offices.  They only have textual descriptions, so we'd probably
 use the "summary" field of each Place object to include the description
 of the place.

 That alternative would be more complicated to implement, however, since
 the form choices would have to be built from the database; if it were
 just the constituencies in the drop-down this would be easy, but to
 group them by county would mean a ModelForm with a queryset that
 finds areas of type CON (the constituency MapIt type) grouped by the DIS
 (county MapIt type) with which they have most overlap. (The counties
 don't exactly include the constituencies; the shapefiles they were
 imported from don't match up.)


pombola/kenya/static/sass/_kenya_overrides.scss
```diff
@@ -439,3 +439,26 @@ body {
...
+  input {
+    margin-bottom: 2em;
+    margin-top: 0.5em;
+    font-size: 1em;
+  }
+}
+
+.iebc-office-result {
+  margin-left: 0.5em;
+  maring-right: 0.5m;
```
jacksonj04: Nasty typo in the SCSS.

jacksonj04:
A tentative :+1:, I can't see anything obviously wrong (apart from the typo in the SCSS), but I don't have a populated MapIt instance to hand to sanity check that things are matching up as expected.

pombola/kenya/static/sass/_kenya_overrides.scss
mhl: Oh, good spot - thanks.

Issue #2183, opened 2016-07-18
mirror / cache external API results to cope with network problems and improve response times


mhl:
The People's Assembly site has gone down a number of times recently. We believe this is because of pages that depend on the results from external APIs; when network connectivity is bad or there are problems at the remote end, these requests can get stuck. We tried to address this with f3e881aba73eff5, which added a 3.05 second timeout to each request. However, this seems not to have fixed the problem - perhaps the timeout isn't working on certain types of error, or when you get a lot of requests the site's effectively always waiting for those to time out.

In any case, the site is still becoming unavailable for end users and this is causing bother for people on the on-call rota, so we should work on making this more robust. This ticket is to represent that work - for example, one implementation *might* look like:

* A management command is run from cron each night that populates the Django cache with up-to-date data with a 5 day timeout (since sometimes it has taken days for the external services to come back). This should report errors if the services are down so we can investigate and prompt the upstream data providers.
* Requests are fulfilled from the cache (this already happens, in fact)

Examples of the external HTTP-based API requests that are made at the moment include:

* http://nearby.code4sa.org/councillor/ward-52401006.json
* https://api.pmg.org.za/member/121/attendance/
* http://api.pmg.org.za/committee-meeting-attendance/data.xlsx
* https://api.pmg.org.za/member/?filter%5Bpa_link%5D=http%3A//www.pa.org.za/person/beverley-lynette-abrahams/
* http://mapit.code4sa.org/point/4326/18.423889,-33.925278?type=WD

PR #2184, opened 2016-07-19
WIP: mirror external HTTP-based API data

mhl:
This is only still work-in-progress because `nearby.code4sa.org` is down at the moment (e.g. http://nearby.code4sa.org/councillor/ward-52401006.json is 404) - this is usually due to the data source upstream of Code4SA's API (the IEC API) being down. The remaining things to be done are:
- [ ] Test the mirroring scripts and updated views on staging
- [ ] Deploy to production in two stages:
  - [ ] Deploy everything but the view changes
  - [ ] Run both `south_africa_populate_api_caches` and `south_africa_mirror_mapit_wards`
  - [ ] Deploy the changes to the views as well.

For #2183

### DIFF
--- a/pombola/south_africa/management/commands/south_africa_mirror_mapit_wards.py
+++ b/pombola/south_africa/management/commands/south_africa_mirror_mapit_wards.py
@@ -1,0 +1,64 @@
+from contextlib import closing
+import os
+from tempfile import NamedTemporaryFile
+import time
+from urlparse import urljoin
+
+import requests
+
+from django.contrib.gis.gdal import DataSource
+from django.core.management.base import BaseCommand
+from django.db import transaction
+
+from mapit.models import Area, Country, Geometry, Generation, Type
+from mapit.management.command_utils import save_polygons
+
+SLEEP_BETWEEN_REQUESTS = 1
+
+class Command(BaseCommand):
+
+    def handle(self, *args, **options):
+        remote_mapit_url = 'http://mapit.code4sa.org/'
+        all_wards_url = urljoin(remote_mapit_url, '/areas/WD')
+        r = requests.get(all_wards_url)
+        time.sleep(SLEEP_BETWEEN_REQUESTS)
+        with transaction.atomic():
+            # Remove existing wards from MapIt:
+            Geometry.objects.filter(area__type__code='WD').delete()
+            Area.objects.filter(type__code='WD').delete()
+            # Get the Country object for South Africa:
+            country, _ = Country.objects.get_or_create(
+                name='South Africa',
+                defaults={'code': 'ZA'})
+            # Get the current Generation:
+            generation = Generation.objects.current()
+            # Make sure the ward type exists:
+            ward_type, _ = Type.objects.get_or_create(
+                code='WD', defaults={'description': 'Ward'})
+            for wd_id, wd_data in r.json().items():
+                # FIXME: we don't actually need the extra names (from
+                # 'all_names'), the codes (from 'codes') or the
+                # parent_area, but use the other attributes. We might
+                # want to add them later.
+                area = Area.objects.create(
+                    country=country,
+                    generation_low=generation,
+                    generation_high=generation,
+                    name=wd_data['name'],
+                    type=ward_type,
+                )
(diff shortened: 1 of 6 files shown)


mhl:
A note since I'm going off on holiday - rather than setting a 5 day time-to-live on data added to the cache, it's probably better to make it infinite. The downtime of the ward councillor data does seem to be very long sometimes, and Code4SA have typically let us know when we should clear the cache because it might contain invalid or stale data.

pombola/south_africa/management/commands/south_africa_mirror_mapit_wards.py
```diff
@@ -0,0 +1,64 @@
...
+    def handle(self, *args, **options):
+        remote_mapit_url = 'http://mapit.code4sa.org/'
+        all_wards_url = urljoin(remote_mapit_url, '/areas/WD')
+        r = requests.get(all_wards_url)
+        time.sleep(SLEEP_BETWEEN_REQUESTS)
+        with transaction.atomic():
+            # Remove existing wards from MapIt:
+            Geometry.objects.filter(area__type__code='WD').delete()
+            Area.objects.filter(type__code='WD').delete()
+            # Get the Country object for South Africa:
```
struan: It might be useful to have an option to not do this and just download missing ones for occasions where the import fails half way through.

pombola/south_africa/management/commands/south_africa_populate_api_caches.py
```diff
@@ -0,0 +1,48 @@
...
+                person = get_person_from_pa_url(member_data['pa_link'])
+                update_or_create_pmg_api_identifier(person, member_data['id'])
+            if not data['next']:
+                break
+            page += 1
+        # Now remove that Identifier from any people that it's disappeared from:
+        Identifier.objects.filter(
+            content_type=ContentType.objects.get_for_model(Person),
+            scheme='za.org.pmg.api/member',
+        ).exclude(object_id__in=pmg_member_ids).delete()
```
struan: I don't see where `pmg_member_ids` is populate for this to work.

pombola/south_africa/views.py
```diff
@@ -162,31 +163,14 @@ def get_object(self):
...
-        url_fmt = 'http://nearby.code4sa.org/councillor/ward-{ward_id}.json'
-        try:
-            r = requests.get(
-                url_fmt.format(ward_id=ward_id),
-                timeout=API_REQUESTS_TIMEOUT)
-            r.raise_for_status()
-            ward_result = r.json()
-        except (requests.exceptions.RequestException, ValueError) as e:
-            raise WardCouncillorAPIDown(unicode(e))
-        councillor_data = ward_result['councillor']
```
struan: It looks like this still needs to happen once we've fetched the data from the cache.

struan: Although it looks like `ward_result` is used further down (line 190) so possibly this still needs to be the two step process.

struan: Similarly `mapit_json` and 'ward_id` are used a bit further down the file.

struan:
A few small things to fix but mostly this seems to work and looks good and sensible.

pombola/south_africa/management/commands/south_africa_mirror_mapit_wards.py
mhl: It's in a `transaction.atomic`, so you should get a complete update or nothing at all.

On a related matter, though, I'm a bit uneasy about removing everything and recreating, just in case someone tries to create foreign keys to areas of type `WD`, since the cascading delete behaviour will remove the other objects. I don't know if "don't do that then" (since they'll discover pretty quickly) is a good enough reponse to this. (Even if we changed the code to only add / remove differences, there's still the risk that someone pointing foreign keys at areas of type `WD` will have objects deleted unexpectedly if they disappear from the code4sa mapit - at least deleting all and recreating will make it really obvious.)

chrismytton:
This pull request hasn't seen much activity for a couple of years and has a few conflicts, so I'm going to close it. Please do feel free to reopen if you think it's useful!

PR #2185, opened 2016-07-25
Add treatment pages for Mzalendo political awareness survey

jacksonj04:
As part of a study on political engagement in Kenya, we need to expose people to two treatment pages and track their interaction.

Following taking part in [a survey](http://www.surveygizmo.com/s3/2802241/34d7088db686/SG_TEST_RUN), users will be directed to one of two pages, one detailing Kenya's involvement in the Olympics, the other detailing the Kenyan political system.

screen shot 2016-07-25 at 14 58 06

screen shot 2016-07-25 at 14 58 11

These pages have a few interactions possible:

screen shot 2016-07-25 at 15 13 23

screen shot 2016-07-25 at 15 13 12

As part of this interaction, the pages will make AJAX calls back to Pombola (more specifically the views in `views_facebook_experiments.py`), which will then (in theory) hit up the SurveyGizmo API to record those facts.


pombola/kenya/static/js/facebook-experiment.js
```diff
@@ -0,0 +1,170 @@
...
+}
+
+$(function(){
+
+    // Show the hidden page element (we don't show anything if no JS is available)
+    $('#ctas').show();
+
+    var options = {
+            placeholder: 'Constituency',
+            allowClear: true,
```
zarino: Last item in javascript object shouldn't end with a comma.

pombola/kenya/static/js/facebook-experiment.js
```diff
@@ -0,0 +1,170 @@
...
+        },
+
+        modelMatcher = function (term, text) {
```
zarino: Indentation inside this `modelMatcher` class is all over the place.

pombola/kenya/static/js/facebook-experiment.js
```diff
@@ -0,0 +1,170 @@
...
+            data: {
+                area: iebcSelector.val()
+            },
+            type: "GET",
+            dataType: "html",
+            success: function (data) {
+                $('#constituency-lookup-result').html(data);
+            },
+            error: function (xhr, status) {
+                $('#constituency-lookup-result').html('<p>Sorry, something went wrong!</p>');
```
zarino: Are we happy with this error message? If it does end up getting shown to people, it won’t be very helpful.

zarino:
The HTML and SCSS bits of this look good to me. Just a few comments left, regarding syntax and an ajax error message.

You’ll need to find someone like @chrismytton to review the Django bits.

pombola/kenya/static/js/facebook-experiment.js
mhl: The code in `modelMatcher` is also a duplicate of the definition in `pombola/kenya/templates/iebc_office_single_select.html`. Instead it would be good to put that into a new `.js` file, which is included both in the 'facebook-experiment` group and a new group for the `iebc_office_single_select.html`. (It should have "Select2" somewhere in the function name too.)

pombola/kenya/static/js/facebook-experiment.js
```diff
@@ -0,0 +1,170 @@
...
+            }
+        });
+
+        // Do we have a SID?
+        if (sid = getParameterByName('sid')) {
```
mhl: Is `sid` meant to be a global - I'm assuming not, so there should be a `var sid` at the start of this function.

pombola/kenya/views_facebook_experiments.py
```diff
@@ -0,0 +1,200 @@
...
+        question_id = request.GET.get('qid')
+        value = request.GET.get('value')
+
+        if survey_session and question_id and value:
+            recordResponseValue(survey_session, question_id, value)
+
+        return super(SurveyGizmoValue, self).get(request, *args, **kwargs)
+
+
+def getSurveyIdFromSession(survey_session):
```
mhl: Functions in Python should be lower case, separated by underscores. This function is only used in this module, so ideally you should prefix it with an underscore too. The same applies to the other functions defined below.

pombola/kenya/views_facebook_experiments.py
```diff
@@ -0,0 +1,200 @@
...
+            "filter[field][0]": "[question(13)]",
+            "filter[operator][0]": "=",
+            "filter[value][0]": survey_session
+            }
+
+        headers = {
+            'cache-control': "no-cache"
+            }
+
+        response = requests.request("GET", url,
```
mhl: Is there a reason to use `requests.request("GET"` rather than `requests.get(` here? (Similarly with the POST below.)

pombola/kenya/views_facebook_experiments.py
```diff
@@ -0,0 +1,200 @@
...
+
+        url = "https://restapi.surveygizmo.com/v4/survey/2802241/surveyresponse/"
+
+        querystring = {
+            "api_token": settings.SURVEYGIZMO_API_TOKEN,
+            "api_token_secret": settings.SURVEYGIZMO_API_SECRET,
+            "filter[field][0]": "[question(13)]",
+            "filter[operator][0]": "=",
+            "filter[value][0]": survey_session
+            }
```
mhl: Python in Pombola either uses one of these two conventions for the trailing `}` of dictionary literals, so it would be best to stick to one of them:

```
    foo = {
        'a': 'b',
        'c': 'd'
    }
```

```
    foo = {
        'a': 'b',
        'c': 'd'}
```

pombola/kenya/views_facebook_experiments.py
```diff
@@ -0,0 +1,200 @@
...
+
+        headers = {
+            'cache-control': "no-cache"
+            }
+
+        response = requests.request("GET", url,
+                                    headers=headers,
+                                    params=querystring
+                                    )
+        data = json.loads(response.text)
```
mhl: `data = response.json()` is a bit simpler.

pombola/kenya/views_facebook_experiments.py
```diff
@@ -0,0 +1,200 @@
...
+
+        # Sanity check we only have one response. If not, abandon ship.
```
mhl: If the authentication details are missing or wrong, then `data` won't contain a `total_count` key - it'll be something like:

```
{u'message': u'Login failed / Invalid auth token', u'code': 401, u'result_ok': False}
```

Rather than throwing a `KeyError` in this case, it would be better to throw an `Exception` with a message that indicates what's wrong.

pombola/kenya/views_facebook_experiments.py
```diff
@@ -0,0 +1,200 @@
...
+
+class TreatmentOlympics(TemplateView):
+
+    template_name = 'facebook-experiment-olympics.html'
+
+    def get(self, request, *args, **kwargs):
+        self.survey_session = request.GET.get('sid')
+
+        if self.survey_session:
+            recordPositiveResponse(self.survey_session, '62', '10173')
```
mhl: It would be good to define the magic integers here as helpfully named constants at the top of the module.

pombola/kenya/views_facebook_experiments.py
```diff
@@ -0,0 +1,200 @@
...
+
+class TreatmentOlympicsThanks(TemplateView):
+
+    template_name = 'facebook-experiment-olympics-thanks.html'
+
+    def get(self, request, *args, **kwargs):
+        survey_session = request.GET.get('sid')
+
+        if survey_session:
+            recordPositiveResponse(survey_session, '58', '10170')
```
mhl: Just to check, since this is processing a GET request, is recording the survey response idempotent?

pombola/kenya/views_facebook_experiments.py
```diff
@@ -0,0 +1,200 @@
...
+        context['sid'] = self.survey_session
+        return context
+
+
+class TreatmentOlympicsThanks(TemplateView):
+
+    template_name = 'facebook-experiment-olympics-thanks.html'
+
+    def get(self, request, *args, **kwargs):
+        survey_session = request.GET.get('sid')
```
mhl: The `sid` value is untrusted input, but it's used verbatim as part of the request to surveygizmo. At the least, I would check in each case that it's only digits (or only hexadecimals, or whatever it's meant to be!) Also, being in the URL it's very public, so it would be easy for someone to observe and repeat. Is that a problem? Are the IDs guessable?

pombola/kenya/views_facebook_experiments.py
```diff
@@ -0,0 +1,200 @@
...
+        return super(SurveyGizmoValue, self).get(request, *args, **kwargs)
+
+
+def getSurveyIdFromSession(survey_session):
+
+    survey_id = cache.get(survey_session)
+
+    if not survey_id:
+
+        url = "https://restapi.surveygizmo.com/v4/survey/2802241/surveyresponse/"
```
mhl: The URL prefix is used in several places - better to define it in a constant, I think

mhl:
These pages don't work without Javascript, since there's no submit button for the office lookup or olympics drop-downs.

pombola/kenya/views_facebook_experiments.py
```diff
@@ -0,0 +1,200 @@
...
+            "api_token": settings.SURVEYGIZMO_API_TOKEN,
+            "api_token_secret": settings.SURVEYGIZMO_API_SECRET,
+            "data[" + question_id + "][" + selection_id + "]": "Yes"
+            }
+
+        headers = {
+            'cache-control': "no-cache",
+            }
+
+        response = requests.request("POST", url, headers=headers, params=querystring)
```
mhl: `response` in this and the next function is unused; either the `response = ` should be dropped or it should be used to handle error responses (e.g. with `response.raise_for_status()`) - presumably it would be good to get an email if these POSTs are failing.

mhl:
I think that's all my comments for the moment. I can't easily test the SurveyGizmo interactions, of course. Some tests would be good for this (even smoke tests for the main pages, mocking the requests), but since this has been in review for 9 days at least partly due to the timing of it being made when I was just going on holiday, I won't insist on it :)

pombola/kenya/views_facebook_experiments.py
jacksonj04: Yes, it's marking a single value as "true", which we can repeat as many times as we feel like.

jacksonj04:
The calls to action are now correctly hidden on pages by default, and are shown by the JS once jQuery has finished its stuff.

pombola/kenya/templates/facebook-experiment-politics.html
```diff
@@ -129,7 +129,7 @@
     </div>
 </div>
 
-<div id="ctas">
+<div id="ctas" hidden>
```
mhl: IE earlier than 11 doesn't support the `hidden` attribute according to http://caniuse.com/#feat=hidden so maybe the CSS equivalent `display: none` would be better.

PR #2190, opened 2016-08-04
Update Ward Councillor Error Message

### LINKED ISSUE #2187: [ZA] Change Ward Rep locator error message
Currently the data from the API the provides us with Ward Rep info won't return results because ZA is having local gov elections on Aug 3rd and therefore they won't have any data on who is in position for a while.

PMG would like the error message returned when searching for a ward rep using the rep locator to change.

Right now the page returned looks like this:
screen shot 2016-08-02 at 14 52 27

PMG would like the wording to be changed to:
Ward councillor data is currently unavailable due to the elections. We will update this when the data becomes available.

This wording should only be displayed on the results page for the rep locator and should look exactly the same as the current message in design and display.

JenMysoc:
Changed Error message for Ward Councillor results not being returned. Fixes #2187

### DIFF
--- a/pombola/south_africa/templates/south_africa/latlon_local_view.html
+++ b/pombola/south_africa/templates/south_africa/latlon_local_view.html
@@ -15,8 +15,7 @@
 
 {% if ward_data_not_available %}
   <div class="ward-councillor-api-down">
-    Ward councillor data is currently unavailable. Please try
-    again later to find your ward councillor details.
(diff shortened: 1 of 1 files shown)


mhl:
:+1: This branch was created from further back in history from current master, so I've cherry-picked this commit onto master. (I also removed the trailing whitespace from the line, since the history would be changed anyway.)

It's deployed now!

PR #2191, opened 2016-08-04
2186 change newsletter and subscribe location

### LINKED ISSUE #2186: [ZA] Change where the newsletter icon and subscribe link goes to
PMG have asked if we could change where the subscribe link in the footer of the site and the yellow newsletter icon in both the header and footer go to.

screen shot 2016-08-01 at 09 03 37

The new link is https://mail4.marketingautomation.services/prospector/form/MzawMDE2MrUwBwA/SzMyMjE3NErSNTAzSdI1SbFI07VITLTQTTZJM05NSjOzNEy1AAA

The newsletter page subscribe link on http://www.pa.org.za/info/newsletter should also be changed to go to this new link although the link to this newsletter page has been removed from the site. 

screen shot 2016-08-01 at 11 46 36

JenMysoc:
Changing where the subscribe buttons and newsletter icons point on People's Assembly. Fixes #2186

### DIFF
--- a/pombola/south_africa/templates/south_africa/_newsletter_signup.html
+++ b/pombola/south_africa/templates/south_africa/_newsletter_signup.html
@@ -1,9 +1,12 @@
 <h4>Sign up to our mailing list</h4>
 <div class="email-signup">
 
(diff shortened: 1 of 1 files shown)


mhl:
This is on http://za-pombola.staging.mysociety.org/ now if you want to try. The links from the envelope icon in the header and footer work fine, but the form in the footer where you just type in your email address doesn't work; it just takes you to the same form (because you wouldn't expect the the URL to accept the same parameters when POSTed to as the old one did).

They also sent us some Javascript for adding a form to the page - I'd better review that to see if we're happy including that.

JenMysoc:
Ah yes, I see. I'll leave that to you!

Issue #2189, opened 2016-08-02
[ZA] Make it simpler to find an MP by name 


JenMysoc:
Currently if you want to look up an MP by name on People's Assembly you need to go to either People or Hot Topics.

If you go to people, you need to know which House they are a member of, and for the council of provinces you need to know which province they stand for. 

![screen shot 2016-08-02 at 15 49 21](https://cloud.githubusercontent.com/assets/4170351/17333048/c9996a88-58c8-11e6-8385-50cc58fc80be.png)

Then if you choose a house you either have the choice of all members, then searching through an alphabetical list or you need to know what party they are in. 

![screen shot 2016-08-02 at 15 49 37](https://cloud.githubusercontent.com/assets/4170351/17333104/f40ad658-58c8-11e6-8355-db0175bc02d6.png)

If you go via the Hot Topics route you will have an alphabetically ordered list but to search you need to know the province the MP stands for:

![screen shot 2016-08-02 at 15 21 02](https://cloud.githubusercontent.com/assets/4170351/17333155/1c560ea2-58c9-11e6-94a8-6507c9a4774d.png)

You can of course use the search box on the menu bar to search for the name of the MP you want, but PMG have said that this is not intuitive for their users. 

People are expecting to be able to search using a search box on the page they have gone to to look for an MP. Currently we are expecting people to click two to three times before they come up against a list of names where they might look for their MP and we're asking them to know a lot of information. The suggestion is that this user journey could improve. 

The simplest solution would appear to be replicating the search box on the Menu bar on both the page for the National Assembly and the National Council of Provinces:
http://www.pa.org.za/organisation/national-assembly/
http://www.pa.org.za/organisation/ncop/

This search should also be on the All members pages:
http://www.pa.org.za/organisation/national-assembly/people/
http://www.pa.org.za/organisation/ncop/people/

And on: http://www.pa.org.za/position/member/parliament/?order=name

Alternatively we could test adding terms to the search box and a bit more description about what you can search for similar to the search on ShineYourEye.org:

![screen shot 2016-08-02 at 16 01 20](https://cloud.githubusercontent.com/assets/4170351/17333486/859af39a-58ca-11e6-94a8-707873881045.png)

This would be the quickest first pass and we could A/B test or similar to see how effective it is.

PR #2194, opened 2016-08-12
ZA: Make it easier to find MPs

mhl:
This implements Jen's suggested changes in #2189 to make it
easier for user to search for representatives.

screenshot from 2016-08-12 15-13-36

screenshot from 2016-08-12 15-14-30

screenshot from 2016-08-12 15-15-07

screenshot from 2016-08-12 15-15-28

### DIFF
--- a/pombola/south_africa/templates/menu_entries.html
+++ b/pombola/south_africa/templates/menu_entries.html
@@ -16,6 +16,7 @@
 <li><a href="{% url "person_list" %}" class="has-dropdown">People</a>
   {% if include_sub_menu_entries %}
     <ul id="parliament-overview">
+      <li><a href="{% url "core_search" %}?section=persons">Find by name</a></li>
       <li><a href="{% url "organisation_kind" slug="parliament" %}">Parliament</a>
         <ul>
           <li><a href="{% url "organisation" slug="national-assembly" %}">National Assembly</a></li>


pombola/south_africa/static/sass/_south-africa_overrides.scss
```diff
@@ -1737,6 +1737,21 @@ a.alphabet-links__link {
     }
 }
 
+li, div {
+    span.inline-search {
+         padding: 0.33em;
+         display: inline-block;
```
chrismytton: I think these two lines should probably be indented 4 spaces rather than 5, to be consistent with the rest of the file :)

pombola/south_africa/templates/core/organisation_people.html
```diff
@@ -9,6 +9,14 @@
 {% block content %}
 
   <h2>People at {{ object.name }}</h2>
+
+  <div class="people-list-inline-search">
+    <span class="">Search for people:</span>
+    <form class="inline-search" action="{% url "core_search" %}">
+      <input class="search-autocomplete-name" id="core_search" data-source="/search/autocomplete/?model=person" type="text" name="q" value="" placeholder="Name"></input>
```
chrismytton: Do you think it's worth having a hidden input here which restricts the `core_search` results to people?

```html
<input type="hidden" name="section" value="persons">
```

pombola/south_africa/templates/south_africa/organisation_house.html
```diff
@@ -27,6 +27,12 @@
   <div>
     <ul class="house-page-list">
       <li><a href="{% url 'organisation_people' slug=object.slug %}">All members</a></li>
+      <li>
+        <span class="inline-search">Search for a person:</span>
+        <form class="inline-search" action="{% url "core_search" %}">
+          <input class="search-autocomplete-name" id="core_search" data-source="/search/autocomplete/?model=person" type="text" name="q" value="" placeholder="Name"></input>
```
chrismytton: As above, perhaps worth adding a hidden input here as well?

chrismytton:
Couple of comments inline, but overall this is looking really good! ✨  🚀

pombola/south_africa/static/sass/_south-africa_overrides.scss
mhl: Thanks, fixed in 684c8bb83008756

pombola/south_africa/templates/core/organisation_people.html
mhl: Good point, I've done that in: ea15dc92a174

pombola/south_africa/templates/south_africa/organisation_house.html
mhl: Done in the same commit: ea15dc92a174

mhl:
I've added some plain submit buttons, to make sure the search works everywhere. It'd be nice to style them like the magnifying glass in the header search:

screenshot from 2016-08-12 15-23-03

But that can be done if it turns out that this change makes a difference...

zarino:
Fly by comment: Should it be "Search for people" or "Search for a person"?

screen shot 2016-08-12 at 15 33 51

(I have a feeling people will most likely be after a specific person, rather than multiple people matching the same search term.)

Indeed, the more I think about it, the more the field label should define what sort of data the input expects – maybe something like "Find a person by name"? – then the input placeholder would be used to show an example name, like "Joe Bloggs" or whatever.

mhl:
Good point, @zarino, I've changed that, using as the example name Baleka Mbete, who's the speaker of the National Assembly.

PR #2196, opened 2016-08-24
Links generated to a sitting should actually show that sitting

### LINKED ISSUE #2195: [KE] All Appearances for Johnson Arthur Sakaja linking to wrong place
When you go to Johnson Arthur Sakaja's profile on Mzalendo and click on any of the parliamentary appearances you are taken straight to the hansard record of Cyprian Kubai Iringo from July 6th 2016. 

screen shot 2016-08-15 at 09 56 24

screen shot 2016-08-15 at 10 43 36

Cyprian's profile is showing the correct speeches and when you click through on any appearance it matches the right date and appearance in hansard. So this only appears to affect Johnson Arthur Sakaja's profile.

mhl:
Sitting.get_absolute_url would omit the start_time suffix if
start_time was None, but the SittingView would interpret that
URL as referring arbitrarily to the first sitting in that venue
on that date.  This meant that appearances might link to the
wrong sitting, confusingly.

This commit changes that logic so that if the start_time is
missing, it looks for a sitting with start_time None.

Fixes #2195

### DIFF
--- a/pombola/hansard/views.py
+++ b/pombola/hansard/views.py
@@ -60,6 +60,36 @@ class IndexView(TemplateView):
         return context
 
 
+def get_sittings_from_slugs(venue_slug, start_date_and_time):
+
+    query_args = {
+        'venue__slug':     venue_slug,
+    }
+
+    # add blank HH-MM-SS if needed
+    start_time_parts = re.split('-', start_date_and_time)
+    if len(start_time_parts) == 3:
+        start_time_parts.extend( [ '00','00','00' ] )
+
+    # split the date and time up into the relevant fields
+    (year,month,day,hour,minute,second) = [ int(x) for x in start_time_parts ]
+
+    # add start_date to the query
+    query_args['start_date'] = datetime.date(year=year,month=month,day=day)
+
+    # if there is a time use it
+    if hour or minute or second:
+        query_args['start_time__gte'] = datetime.time(hour=hour, minute=minute, second=second)
+
+    # get all matching sittings
+    sittings = Sitting.objects.filter( **query_args ).order_by('start_time')
+
+    if not len(sittings):
+        raise Http404
+
+    return sittings
+
+
 class SittingView(DetailView):
     model = Sitting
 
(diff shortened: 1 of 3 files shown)


pombola/hansard/views.py
```diff
@@ -60,40 +60,48 @@ def get_context_data(self, **kwargs):
...
-        if len(start_time_parts) == 3:
-            start_time_parts.extend( [ '00','00','00' ] )
+    start_date, start_time = m.groups()
+    if start_time:
+        start_time = start_time.replace('-', ':')
 
-        # split the date and time up into the relevant fields
-        (year,month,day,hour,minute,second) = [ int(x) for x in start_time_parts ]
+    query_args['start_date'] = start_date
+    query_args['start_time'] = start_time
```
chrismytton: Is it OK to pass plain strings like this to `.filter` when the actual field is a date? I couldn't see anything in the docs but I might well have just been looking in the wrong place!

chrismytton:
I'm finding this diff a little tricky to follow. I *think* it's because the commit is doing at least three distinct things:

- Extracting some logic from `SittingView.get_object` into `get_sittings_from_slugs`
- Adding a (very useful!) error message to the 404 error
- Fixing the original bug

But I don't think it's worth splitting the commit up into three separate ones, unless you really want to :)

I left one query, just checking that my understanding of Django ORM's date handling is correct. Other than that, this looks good to me! 👍

mhl:
@chrismytton: Oh, you're right, the diff isn't very nice. I've split that into multiple changes that are more readable now.

mhl:
@chrismytton: With regard your other query about using strings instead of `date` objects in the query against the `DateField`, I believe that's allowed, but I can't find any documentation supporting that - maybe it's just that it passes down the strings literally and PostgreSQL is happy with comparing against those strings but some other database might not, so I've asked about that on StackOverflow: http://stackoverflow.com/q/39140914/223092

chrismytton:
Ah, the diff is _much_ easier to follow now! ✨

chrismytton:
> @chrismytton: With regard your other query about using strings instead of date objects in the query against the DateField, I believe that's allowed, but I can't find any documentation supporting that - maybe it's just that it passes down the strings literally and PostgreSQL is happy with comparing against those strings but some other database might not, so I've asked about that on StackOverflow: http://stackoverflow.com/q/39140914/223092

👍  Yep, good shout.

PR #2198, opened 2016-09-07
"About" dropdown in People's Assembly header

### LINKED ISSUE #2192: [ZA] Move link items in the header
PMG have asked if we could change links in the header of the site.

They would like the current "About" link to become a drop down list of links that would include

Who are we

For Elected Representatives

Source Data

The header would then just include About and Contact Us

screen shot 2016-08-02 at 15 20 37

zarino:
Fixes #2192 

screen shot 2016-09-07 at 12 44 51

Also includes refactoring of how dropdowns are handled across all the pombola menus, simplifying the code, and making them more usable on touchscreens and keyboard navigation.

Tested in all the major browsers <del>apart from Opera Mini (can't get it working locally – maybe we can test once this has been deployed to staging?).</del>

I’ve noticed the menu lacks any styling on IE8 – but that’s happening on the live site too! Can't tell what's causing it. Thought maybe it was a missing html5shiv, but I tried adding it and nothing changed.


pombola/core/static/js/hover-dropdown.js
```diff
@@ -0,0 +1,69 @@
+$(function(){
+
+    /*
```
zarino: I wonder whether, if this comment were in single line comment format (`//`) whether it'd get stripped out by the django pipeline? Seems a shame to be sending all this documentation to the client.

zarino: Alternatively, we could remove the comment entirely, and put the documentation into `_hover-dropdown.scss`

zarino:
Houston we have a problem: The "About" dropdown menu isn't triggering on Opera Mini 14.0.0 on my iPhone. Pressing "About" just navigates to the About page. Long pressing brings up the browser’s context menu ("Open in new tab" etc) but doesn't focus the link.

pombola/core/static/js/hover-dropdown.js
mhl: Yep, yui-compressor does remove single line comments, so I'll make that change and fixup that commit.

mhl:
This all looks good to me, other than the Opera Mini problem...

> Maybe we should check for window.operamini and if it’s present, bind our touchstart shim to click as well.

This sounds good to me.

zarino:
@mhl: Opera Mini big fixed now. It's not as elegant as I'd like, due to some unexpected `e.preventDefault()` behaviour, but I've tested it, and this is the one and only incantation that seems to work. Clicking the "About" link reloads the page (unavoidable) and once the page reloads, the dropdown menu is visible, and all the links inside it are clickable.

PR #2201, opened 2016-09-30
[Mzalendo] Update treatment pages for second UCL pilot survey

jacksonj04:


pombola/kenya/static/js/facebook-experiment.js
```diff
@@ -1,3 +1,5 @@
+var logbookUrl = 'http://logbook.mysociety.org/mzalendo-fb';
```
mhl: What's `logbook.mysociety.org`? I can't find anything about it with a search on the wiki for 'logbook' ;)

PR #2203, opened 2016-10-04
pombola_sayit: also create Identifier objects with Pombola slugs

mhl:
We need to add all the possible Pombola Person slugs as Identifier
objects on the SayIt Speaker so that we can use them in name resolution
when processing data from the Code4SA / PMG API for questions and
answers.


struan:
This looks good to me.

PR #2204, opened 2016-10-06
Add a template tag for "cleaning" HTML using lxml.html.clean

mhl:
This should remove HTML that might be dangerous (e.g. script tags)


struan:
Looks good to me.

PR #2218, opened 2016-11-22
Consistent styling for `.people-list-inline-search`

### LINKED ISSUE #2216: [ZA] Increase visibility of the name search box on MP Profiles page
PMG had overlooked the work we did on the search box on the MP profiles page and were also unaware that the box allowed you to start typing and the word would autocomplete. 

screen shot 2016-11-21 at 12 49 03

Above is the search box on the page as it currently looks. Can a designer think of a way to make this more visible and eye catching to a user? http://www.pa.org.za/position/member/parliament/?order=name

zarino:
Fixes #2216.

The search box at the top of the "MP Profiles" page now matches the main search page:

screen shot 2016-11-22 at 16 43 08

### DIFF
--- a/pombola/core/static/sass/_search.scss
+++ b/pombola/core/static/sass/_search.scss
@@ -118,7 +118,15 @@
       width: 25%;
     }
   }
+}
 
+.people-list-inline-search {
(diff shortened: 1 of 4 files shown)


zarino:
@mhl Could you give this a review and merge if happy?

PR #2219, opened 2016-11-23
ZA: Show 'must read' posts in sidebar

### LINKED ISSUE #2213: [ZA] Create "must read" section on RHS menu on blog page
screen shot 2016-11-18 at 11 23 20

Currently the Blog page RHS menu on PA has two sections; Recent Posts and Popular Posts. PMG have indicated that underneath this section they would like a "Must read" section which links out to blogs they have decided are "must read" blogs.

Ideally, to indicate that a blog entry is "Must Read" PMG would be able to do something like add a tag, or tick a box, to add a blog to the must read section. 

They want to have up to 4 blogs in this section at a time - so we may need to find a way that they can easily find all blogs selected as must read in the admin section - so they can easily deselect ones that have been on there for a while. 

We need more feedback from them on the following things:
- Whether they want the "must read" section to be a section in itself on the main menu dropdown from the blog section
- Whether they expect to remove blog posts from this section a regular basis or if they just expect to only show the most recently added "must read" blogs

octopusinvitro:
The posts must be tagged with a tag whose slug is `must-read`
in order to appear there.

This change updates the django-info-pages package so that we
can use `posts_by_tag`.

We also added a new template tag to the core template tags
which allows us to return the value corresponding to a key when it
contains characters that are not allowed in template variable
names like `,`, `.`, `-`, etc. since it is very common to have dashes in
tag slugs.

Fixes #2213

### DIFF
--- a/pombola/core/templatetags/get_from_key.py
+++ b/pombola/core/templatetags/get_from_key.py
@@ -1,0 +1,11 @@
+from django import template
+
+register = template.Library()
+
+
+@register.filter
+def get_from_key(value, arg):
+    try:
+        return value[arg]
+    except KeyError:
+        return ''
--- a/pombola/core/tests/test_templatetags.py
+++ b/pombola/core/tests/test_templatetags.py
@@ -1,5 +1,6 @@
 # coding=UTF-8
 
+from django.template import Context, Template
 from django.test import TestCase
 
 from ..templatetags.breadcrumbs import breadcrumbs, path_element_overrides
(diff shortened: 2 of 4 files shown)


mhl:
I think we should add some tests for the `get_from_key` template filter, since its implementation is slightly unexpected. There are some simple examples of testing a template filter here: https://github.com/ciudadanointeligente/write-it/blob/alpaca/nuntium/tests/date_localization_tests.py#L33-L52

... and we should probably test:

* a dict where the key is present
* a dict where the key is not present
* a class like LazyTagLookup where the key is present
* a class like LazyTagLookup where the key is not present

octopusinvitro:
@mhl feedback addressed :+1:

PR #2223, opened 2016-11-29
South Africa: Add alphabet navigation to MP profile pages

### LINKED ISSUE #2215: [ZA] Make pagination menu for MP Profiles page into Alphabet menu
PMG would like us to change the pagination menu on the MP profiles to an Alphabet menu. 

screen shot 2016-11-21 at 12 49 17

This has been done elsewhere on the site so we think we may be able to reuse that code 

They only want this feature to be extended to the MP profiles section of the website right now.

octopusinvitro:
This PR:

* Adds a new template tag to add a query parameter to a URL in a template
* Moves the logic for showing an alphabet navigation from the South African views.py to the core so that it can be used elsewhere
* Adds a new query parameter to use in pages where we want to show the alphabet navigation `?a=1`. 

In the screenshot, you can see the url with the new query parameter and the MPs sorted by the letter C

alpha

Fixes  #2215

### DIFF
--- a/pombola/core/tests/test_templatetags.py
+++ b/pombola/core/tests/test_templatetags.py
@@ -1,5 +1,6 @@
 # coding=UTF-8
 
+from django.http import QueryDict
 from django.template import Context, Template
 from django.test import TestCase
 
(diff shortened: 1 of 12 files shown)


pombola/core/tests/test_templatetags.py
```diff
@@ -99,3 +100,23 @@ def test_gets_dynamically_generated_value_from_key(self):
 class LazyDictLike(object):
     def __getitem__(self, key):
         return key + key
+
+
+class AddQueryParameterTest(TestCase):
```
mhl: This looks good - I'd just add one more, which tests the behaviour that an existing value for a key would be replaced

octopusinvitro:
@mhl Feedback addressed in the fixup commits :+1:

PR #2226, opened 2016-12-01
South Africa: Split the views file and make it into a package

octopusinvitro:
This PR splits the South African views file into different files and packages them in a `views` folder.

* Classes that seemed to be related were put in the same file.
* While moving the classes, syntax and style errors detected by Flake were corrected (indentation, spaces, etc.).
* There is still code that can be refactored (for example, the repetition in several classes in the code that uses the `Paginator`) although that could be done in a separate PR.

## Notes to merger

Probably #2223 should be merged first and this branch rebased. Then `SAOrganisationDetailSubPeople` should be updated.

## Update

#2223 is merged, so go to comment https://github.com/mysociety/pombola/pull/2226#issuecomment-264890817


pombola/south_africa/views/geolocalization.py
```diff
@@ -0,0 +1,309 @@
...
+    BasePlaceDetailView, PlaceDetailView, PlaceDetailSub)
+from pombola.search.views import GeocoderView
+
+from pombola.south_africa.models import ZAPlace
+
+# In the short term, until we have a list of constituency offices and
+# addresses from DA, let's bundle these together.
+CONSTITUENCY_OFFICE_PLACE_KIND_SLUGS = (
+    'constituency-office',
+    'constituency-area',  # specific to DA party
```
octopusinvitro: If I remove the space I get the flake8 error:
`E262 - Inline comment should start with '# '`
so I'm going to leave this one unchanged if that is OK.

octopusinvitro:
@mhl Feedback addressed :+1: 

The only thing I didn't correct is the blank space after a `#` in comments, since if I remove the space I get the flake8 error:
`E262 - Inline comment should start with '# '`

I also fixed the error:
`E226 missing whitespace around arithmetic operator`
in files:
* `person.py` and 
* `interests.py`. 

For the latter file I did this correction in an additional fixup commit different from the one where I addressed your feedback.

octopusinvitro:
**Update:** Now that #2223 is merged, I rebased the branch to incorporate those changes and modified the relevant commits so that the tests pass in every commit and the flow of commits in this refactoring makes sense.

The class `SAOrganisationDetailSubPeople` was updated with the changes introduced to it in #2223

I left the fixup commits unmerged since they were addressing your previous feedback.

Issue #2221, opened 2016-11-24
[ZA] Small text changes on MP profiles page


JenMysoc:
Delete two bits of text from the MP Profiles page:

Delete any from Member of any Parliament 

Delete Current Position Holders

PR #2228, opened 2016-12-06
ZA: MP Profiles tiny header fix

octopusinvitro:
This PR addresses the first part of #2221, in particular, deleting "any" from the header of the MP Profiles page.

Since this change affects only a particular page in a particular country, a conditional was added to the template that renders the MP Profiles page in the South African site to discriminate this particular page from other position pages.

For this conditional to work properly, a new item was added to the positions context in the core views so that it also sends the organization kind slug (`ok_slug`) to the template.

The tests cover the three branches of the condition.

### DIFF
--- a/pombola/core/views.py
+++ b/pombola/core/views.py
@@ -384,6 +384,8 @@ def position(request, pt_slug, ok_slug=None, o_slug=None):
         possible_sessions = models.ParliamentarySession.objects.order_by('name')
 
     page_title = title.name
+
+    organisation_kind = None
     if o_slug:
         organisation = get_object_or_404(models.Organisation,
                                          slug=o_slug)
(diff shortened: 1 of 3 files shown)


pombola/core/views.py
```diff
@@ -482,6 +482,7 @@ def position(request, pt_slug, ok_slug=None, o_slug=None):
 
     context = {
         'object':     title,
+        'ok_slug':    ok_slug,
```
mhl: I'm not sure this is really worth changing for this pull request, but it might be more helpful for developers in the future if you passed the `OrganisationKind` object, rather than just the slug? (Also, if you called it `organisation_kind`, it'll be easier for people working in the templates to figure out what it's for.)

mhl: (The `OrganisationKind` object is already being fetched in this function, as you'll probably have seen, so there'd be no additional query.)

octopusinvitro:
@mhl feedback addressed :+1: 

* since I had to rebase to edit the commit messages anyway, I rebased the branch to merge those two commits together.
* I left the core one alone since it is not a `ZA` commit and moved it to the top.
* Then I addressed your feedback as fixup commits so that you can see the changes I did.

PR #2230, opened 2016-12-06
[SA] hide obvious or irrelevant positions and places in people lists

### LINKED ISSUE #2222: [ZA] Removal of superfluous words next to MPs for MP profiles list
On NA MPs PMG do NOT want a province to show 
For example:
screen shot 2016-11-21 at 12 49 03
Freddie here should not have a pin with a province showing!
Also
screen shot 2016-11-24 at 15 51 00
Here on the All members list you can see some MPs have a province in brackets - PMG would like these removed.

This affects MPs on the following pages: 
http://www.pa.org.za/organisation/national-assembly/people/
http://www.pa.org.za/organisation/national-assembly/party/anc/
http://www.pa.org.za/organisation/national-assembly/party/da/
http://www.pa.org.za/organisation/national-assembly/party/eff/
http://www.pa.org.za/organisation/national-assembly/party/ifp/
http://www.pa.org.za/organisation/national-assembly/party/nfp/
http://www.pa.org.za/organisation/national-assembly/party/ff/
http://www.pa.org.za/organisation/national-assembly/party/udm/
http://www.pa.org.za/position/member/parliament/?order=name

PMG would also like the word "member" removed from all NA person lists. The want to keep titles like "Chief Whip"

For NCOP members PMG *would* like the province to show but would like the word Delegate removed. They want to keep titles like "Chief Whip"

screen shot 2016-11-24 at 15 47 08

This affects NCOP people on the following pages:
http://www.pa.org.za/organisation/ncop/people/
http://www.pa.org.za/organisation/ncop/party/anc/
http://www.pa.org.za/organisation/ncop/party/da/
http://www.pa.org.za/organisation/ncop/party/eff/
http://www.pa.org.za/organisation/ncop/party/ifp/
http://www.pa.org.za/organisation/ncop/party/udm/

struan:
This hides place names and positions in people lists where they are either obvious (e.g. all Members of Parliament will have the position Member) or irrelevant (the province of an MP is not very useful).

It does this using template tags to check if the position or place is relevant for the organisation that is currently being displayed.

Fixes #2222

### DIFF
--- a/pombola/core/views.py
+++ b/pombola/core/views.py
@@ -384,7 +384,7 @@ def position(request, pt_slug, ok_slug=None, o_slug=None):
         possible_sessions = models.ParliamentarySession.objects.order_by('name')
 
     page_title = title.name
-
+    organisation = None
     organisation_kind = None
     if o_slug:
         organisation = get_object_or_404(models.Organisation,
@@ -486,14 +486,13 @@ def position(request, pt_slug, ok_slug=None, o_slug=None):
         'object':     title,
         'page_title': page_title,
         'order':      request.GET.get('order'),
+        'organisation_kind':   organisation_kind,
+        'org':        organisation,
         'places':     places,
         'place_slug': place_slug,
         'session': session,
         'session_details': session_details,
     }
-
-    if organisation_kind is not None:
-        context['organisation_kind'] = organisation_kind
 
     if request.GET.get('a') == '1':
         positions, extra_context = filter_by_alphabet(
--- a/pombola/south_africa/templates/core/organisation_party.html
+++ b/pombola/south_africa/templates/core/organisation_party.html
@@ -3,6 +3,7 @@
 {% load staticfiles %}
 {% load thumbnail %}
 {% load hidden %}
+{% load za_people_display %}
 
 
 {% block title %}{{ object.name }} {{ party.slug.upper }} members{% endblock %}
@@ -60,11 +61,17 @@
                   {% endmaybehidden %}
 
                   {% for position in positions %}
-                      {{ position.title }}
-                      {% if position.place %}
+                      {% should_display_position position.organisation position.title as display_position %}
+                      {% if display_position %}
+                        {{ position.title }}
+                      {% endif %}
+                      {% should_display_place position.organisation as display_place %}
+                      {% if display_place and position.place %}
                           (<a href="{{ position.place.get_absolute_url }}">{{ position.place.name }}</a>)
                       {% endif %}
-                      <br/>
+                      {% if display_position or display_place and position.place %}
+                        <br/>
+                      {% endif %}
                   {% endfor %}
 
                 </li>
--- a/pombola/south_africa/templates/core/organisation_party.html
+++ b/pombola/south_africa/templates/core/organisation_party.html
@@ -3,7 +3,6 @@
 {% load staticfiles %}
 {% load thumbnail %}
 {% load hidden %}
-{% load za_people_display %}
 
 
 {% block title %}{{ object.name }} {{ party.slug.upper }} members{% endblock %}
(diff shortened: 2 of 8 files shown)


pombola/south_africa/templatetags/za_people_display.py
```diff
@@ -0,0 +1,29 @@
...
+
+register = template.Library()
+
+NO_PLACE_ORGS = ('parliament', 'national-assembly', )
+MEMBER_ORGS = ('parliament', 'national-assembly', )
+
+
+@register.assignment_tag()
+def should_display_place(organisation):
+    should_display = True
```
mhl: I think it'd be easier to spot potential problems in this implementation if it were just the one-liner `return organisation.slug not in NO_PLACE_ORGS`

pombola/south_africa/templatetags/za_people_display.py
```diff
@@ -0,0 +1,29 @@
...
+
+
+@register.assignment_tag()
+def should_display_position(organisation, position):
+    should_display = True
+
+    if organisation.slug in MEMBER_ORGS and str(position) in ('Member',):
+        should_display = False
+
+    if 'ncop' in organisation.slug and str(position) in ('Delegate',):
```
mhl: What is the `'ncop' in organisation.slug` trying to catch as well as the exact slug `ncop`? (If there is other stuff, I'd add a comment about what it's trying to match, and possibly make it a regexp match so it's more precise - if there aren't other slugs, I'd prefer an `==` match.

I'd make the second part of both conditionals here: `unicode(position) in (u'Delegate',)` with a similar change above in case there are non-ASCII characters in the position title's name. (I don't think it'll fail in any bad way in practice at the moment, but it might help someone in future changing this code.)

pombola/south_africa/templatetags/za_people_display.py
```diff
@@ -0,0 +1,29 @@
...
+    should_display = True
+
+    if organisation.slug in NO_PLACE_ORGS:
+        should_display = False
+
+    return should_display
+
+
+@register.assignment_tag()
+def should_display_position(organisation, position):
```
mhl: I think the second parameter here is a position title, not a position, so renaming the variable accordingly would be good.

pombola/core/views.py
```diff
@@ -486,6 +486,8 @@ def position(request, pt_slug, ok_slug=None, o_slug=None):
         'object':     title,
         'page_title': page_title,
         'order':      request.GET.get('order'),
+        'org_kind':   organisation_kind,
```
mhl: `organisation_kind` is already added to the template in the code a bit further down (although I should have pointed out when reviewing that earlier addition that it should just be included in the dictionary literal, as you've done. oops.)

I'd be tempted to make a commit that moves the setting of `organisation_kind` into the dictionary literal first, then a variant of this commit that calls it `organisation_kind`. (Or have another commit to rename `organisation_kind` to `org_kind` - I don't mind about the naming here.)

BTW, we're deliberately not lining up dictionary values in literals in newly introduced code, since it's a PEP8 violation, in case keeping a minimal diff was part of the motivation for the shorter names.

pombola/south_africa/templates/core/organisation_party.html
```diff
@@ -60,11 +61,17 @@
...
+                      {% should_display_position position.organisation position.title as display_position %}
+                      {% if display_position %}
+                        {{ position.title }}
+                      {% endif %}
+                      {% should_display_place position.organisation as display_place %}
+                      {% if display_place and position.place %}
                           (<a href="{{ position.place.get_absolute_url }}">{{ position.place.name }}</a>)
                       {% endif %}
-                      <br/>
+                      {% if display_position or display_place and position.place %}
```
mhl: I'd be tempted to put a comment above (since I had to look this up) quoting https://docs.djangoproject.com/en/1.8/ref/templates/builtins/#boolean-operators "Use of both and and or clauses within the same tag is allowed, with and having higher precedence than or e.g.:" and explaining that parentheses for precedence aren't allowed.

Or I dunno. Then I suppose you should add the same to the other file - maybe it's distracting noise. Up to you, feel free to ignore this :)

PR #2236, opened 2017-01-30
turn on popolo export for SYE in cron

struan:


conf/crontab.ugly
```diff
@@ -60,6 +60,9 @@ MAILTO=cron-!!(*= $site *)!!@mysociety.org
 !!(* } elsif ($vhost eq 'nigeria.mzalendo.mysociety.org') { *)!!
```
mhl: I think this is an old vhost which has now been removed, the condition should now be based on the vhost `www.shineyoureye.org` - it looks like this means that the feedback reports haven't been happening for ShineYourEye (oops :frowning_face:)

struan: Fixed now.

conf/crontab.ugly
```diff
@@ -57,7 +57,7 @@ MAILTO=cron-!!(*= $site *)!!@mysociety.org
 # Check for any recurring data problems:
 47 4 * * * !!(*= $user *)!! run_management_command kenya_check_for_recurring_problems
 
-!!(* } elsif ($vhost eq 'nigeria.mzalendo.mysociety.org') { *)!!
+!!(* } elsif ($vhost eq 'www.shineyoureye.com') { *)!!
```
mhl: It's `.org` :)

struan: Which is amazing as I got it correct two lines later :(

Fixed now. I will squash all this embarrassment down before I merge :|

mhl:
:+1: (assuming you're going to squash these ;))

PR #2237, opened 2017-02-03
ZA - remove place filter on MP profiles list page

### LINKED ISSUE #2233: [ZA] Remove the dropdown search from the MP profiles page
Now we've implemented a bigger and better search box on the MP profiles page we don't need the drop down search by area on the green bar.

Could we remove this please? 

screen shot 2017-01-17 at 10 45 19

struan:
This removes the dropdown for only showing MPs by area as it's largely been made redundant by the improved search.

Fixes #2233

### DIFF
--- a/pombola/south_africa/templates/core/position_position_section.html
+++ b/pombola/south_africa/templates/core/position_position_section.html
@@ -1,0 +1,9 @@
+{% load pagination_tags %}
+
+{% autopaginate positions %}
+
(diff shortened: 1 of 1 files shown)


mhl:
(BTW, when I said it wasn't likely to be a surprise that I'm commenting on commit messages, that was alluding to the fact that I'm generally picky about them, nothing particular to you!)

mhl:
:+1: Great, thanks!

PR #2238, opened 2017-02-06
Editing message about what attendance is shown

JenMysoc:
As per PMG's request  -have edited the message at the top of the MP Attendance page to simply show that access is restricted to MPs from larger parties.


mhl:
This was successfully merged and deployed, I just managed to confuse GitHub :/

PR #2244, opened 2017-02-09
MPs and Ministers are mutually exclusive

xybrnet:
This change excludes Ministers from the list when viewing the attendance of MPs.

If we cannot determine the `Position` of a `Person` (when the Members returned from the PMG API  has `pa_url=None`), we assume them to be an MP.

The average attendance for parties should only be displayed when viewing MP attendance, and not when viewing the attendance for Ministers, as their attendance is only encouraged and not  compulsory.

### DIFF
--- a/pombola/south_africa/views/attendance.py
+++ b/pombola/south_africa/views/attendance.py
@@ -3,6 +3,7 @@ from __future__ import division
 import dateutil
 import json
 import requests
+import datetime
 
 from .constants import API_REQUESTS_TIMEOUT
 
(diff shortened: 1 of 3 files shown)


pombola/south_africa/views/attendance.py
```diff
@@ -53,20 +54,30 @@ def filter_attendance(self, annual_attendance, party, position):
...
+        year_end_date = datetime.datetime.strptime(
+            annual_attendance['end_date'], "%Y-%m-%d").date()
+        active_ministers = Position.objects.filter(
                 Q(title__slug__startswith='minister') |
                 Q(title__slug__startswith='deputy-minister')
-            ).overlapping_dates(
-                annual_attendance['start_date'],
-                annual_attendance['end_date'])
+            ).currently_active(
+                year_end_date)
```
mhl: I've merged this anyway, because it sounds like it's time-sensitive and you might not be able to look at feedback tomorrow. But anyway, you could simplify this a little bit by doing:

```python
active_minister_slugs = set(Position.objects.filter(
    Q(title__slug__startswith='minister') |
￼    Q(title__slug__startswith='deputy-minister')).currently_active(year_end_date) \
        .values_list('person__slug', flat=True))
```

xybrnet: Great, thank you for pointing that out. There's going to be a couple of tweaks in the next few days, and I'll simplify it with those changes.

PR #2246, opened 2017-02-15
WIP: add NCOP members to the MP Profiles page as well

mhl:
See #2214 - unstyled, this looks like:

draft-new-view

Hopefully this is enough for @zarino to get started with.


zarino:
@mhl 0390dfda8e08af7171461ba17c00f230bc603c1a added styling for `.position-house` – which it turns out we’re no longer using anywhere – do you think I should remove it, or leave it in, in case it comes in useful in future?

PR #2248, opened 2017-02-24
2247 minister attendance

xybrnet:
This enables us to show an attendance of 0 for ministers without attendance records, and to only display a subset of columns for the ministers.

We get all the active Minister Positions from the DB as at the date of the query, or the year end. Whichever is earlier.

For the People associated with active ministerial Positions, for whom we didn't receive data from the PMG API, we add attendance records of 0. If we cannot determine a member in the API results to be a Minister, we assume them to be an MP.

We only show meetings attended for ministers, as attendance is not compulsory, and we don't want to give this impression by showing a ratio.

The filters and disclaimers have also been re-arranged on the page.

### DIFF
--- a/pombola/south_africa/static/js/attendance-table.js
+++ b/pombola/south_africa/static/js/attendance-table.js
@@ -1,8 +1,13 @@
 jQuery (function($) {
   $(document).ready(function(){
+      var order_by;
+      if (POSITION == 'ministers') {
+        order_by = [2, "desc"]
+      }
       $('#mp-attendance').DataTable({
         "paging": false,
-        "info": false
+        "info": false,
+        "order": order_by
       });
   });
 });
--- a/pombola/south_africa/templates/south_africa/mp_attendance.html
+++ b/pombola/south_africa/templates/south_africa/mp_attendance.html
@@ -67,7 +67,7 @@
 
   {% if party %}
    <div class="col-50">
-    {% if position = 'mps' %}
+    {% if position == 'mps' %}
       {% if aggregate_attendance >= 0 %}
         <h2 class="average-attendance">{{ aggregate_attendance }}%</h2>
         <p>Average attendance for {{ party }}</p>
@@ -91,9 +91,11 @@
       <th>Name</th>
       <th>Party</th>
       <th>Meetings</th>
+      {% if position == 'mps' %}
       <th>Present</th>
       <th>Arrived late</th>
       <th>Departed Early</th>
+      {% endif %}
     </tr>
   </thead>
   <tbody>
@@ -101,6 +103,9 @@
     <tr>
       <td><a href="{{ member.pa_url }}">{{ member.name }}</td>
       <td>{{ member.party_name }}</td>
+      {% if position == 'ministers' %}
+      <td>{{ member.present }}</td>
+      {% else %}
       <td>{{ member.total }}</td>
       <td width="15%">
         <div class="attendance">
@@ -127,6 +132,7 @@
           </div>
         </div>
       </td>
+      {% endif %}
     </tr>
    {% endfor %}
   </tbody>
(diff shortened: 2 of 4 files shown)


pombola/south_africa/templates/south_africa/mp_attendance.html
```diff
@@ -5,6 +5,9 @@
 
 {% block js_end_of_body %}
   {{ block.super }}
+  <script type="text/javascript">
+    var POSITION = '{{ position }}';
```
mhl: It doesn't actually matter in this case, since you know that `position` won't have unusual characters in it, but in case someone changes the view code or copy-and-pastes this to somewhere where it isn't safe, I'd make it: `position|escapejs`.

pombola/south_africa/static/js/attendance-table.js
```diff
@@ -1,8 +1,13 @@
 jQuery (function($) {
   $(document).ready(function(){
+      var order_by;
+      if (POSITION == 'ministers') {
+        var order_by = [2, "desc"]
```
mhl: A tiny nitpick, but there `var` on this line isn't necessary, since you've put the `var` at the top of the function anyway, in nice Crockford-style :)

pombola/south_africa/views/attendance.py
```diff
@@ -81,21 +81,24 @@ def filter_attendance(self, annual_attendance, party, position):
...
-
-                    minister_attendance.append({
-                        'member': {
-                            'name': minister_name,
-                            'pa_url': pa_url + minister.get_absolute_url(),
-                            'party_name': minister.parties()[0].slug.upper()},
-                        'attendance': {
-                            'P': 0}
-                    })
+                if party and minister.parties()[0].slug.upper() != party:
```
mhl: Sorry, I don't understand this - could you add a comment in the code explaining why the minister is skipped in this case?

pombola/south_africa/templates/south_africa/mp_attendance.html
xybrnet: Yes, definitely. Thank you.

pombola/south_africa/static/js/attendance-table.js
xybrnet: Nitpicking is good : )

pombola/south_africa/templates/south_africa/mp_attendance.html
```diff
@@ -91,16 +91,21 @@ <h2 class="average-attendance">{{ aggregate_attendance }}%</h2>
       <th>Name</th>
       <th>Party</th>
       <th>Meetings</th>
+      {% if position = 'mps' %}
```
xybrnet: I think this was changed later, to what you are asking here in the second point. Could you perhaps just check and see if it's okay?

pombola/south_africa/views/attendance.py
```diff
@@ -49,36 +49,62 @@ def filter_attendance(self, annual_attendance, party, position):
...
+                if attendance['member']['pa_url'].split('/')[-2] in active_minister_slugs:
+                    minister_attendance.append(attendance)
+                    attendance_summary.remove(attendance)
+                    active_minister_slugs.remove(attendance['member']['pa_url'].split('/')[-2])
 
         if position == 'ministers':
-            attendance_summary = minister_attendance
+            # Some Ministers don't have attendance records,
+            # and we need to show that they haven't attendended any meetings
+            pa_url = 'http://www.pa.org.za'
```
xybrnet: Ah, yes. That's much better doing it this way around. It was for consistency that I prepended it.

xybrnet:
I've made the changes requested above. I think the code could be simpler, so I'm happy for any suggestions.

pombola/south_africa/views/attendance.py
xybrnet: I've added it as a class function. Is that fine?

PR #2252, opened 2017-03-02
Only show no of meetings attended for ministers on Person profile

xybrnet:
This makes the attendance records on Person profiles consistent with the attendance listing page.

For an active ministerial position in a given year, we only display the number of meetings attended, and not a percentage

### DIFF
--- a/pombola/south_africa/views/person.py
+++ b/pombola/south_africa/views/person.py
@@ -248,20 +248,15 @@ class SAPersonDetail(PersonSpeakerMappingsMixin, PersonDetail):
         # Check if the Person held an active ministerial position the last day
         # of the year. Return 'minister' if so, 'mp' if not.
 
-        year_end_date = datetime.date(year, 12, 31)
-        today = datetime.date.today()
-
-        active_minister = self.object.position_set.filter(
-                Q(title__slug__startswith='minister') |
-                Q(title__slug__startswith='deputy-minister')
-            ).currently_active(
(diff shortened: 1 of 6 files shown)


pombola/south_africa/views/person.py
```diff
@@ -243,6 +244,25 @@ def get_latest_meeting_urls(self, data):
...
+        # of the year. Return 'minister' if so, 'mp' if not.
+
+        year_end_date = datetime.date(year, 12, 31)
+        today = datetime.date.today()
+
+        active_minister = self.object.position_set.filter(
+                Q(title__slug__startswith='minister') |
+                Q(title__slug__startswith='deputy-minister')
+            ).currently_active(
+                today if year_end_date > today else year_end_date)
```
mhl: Since this is similar to filtering logic used in your previous filter request, I think it might be good to factor these out into new methods on `pombola.core.models.PositionQuerySet`. e.g. you could have:

```python
    def at_end_of_year(self, year):
        ...
```

... and:

```python
    def title_slug_prefixes(self, title_slug_prefixes):
        qs = [Q(title_slug__startswith=prefix) for prefix in title_slug_prefixes]
        return self.filter(reduce(lambda acc, item: acc | item, qs)
```

(or something like that - I haven't tested that) and chain those methods. What do you think about that?

xybrnet: I think this is pretty cool. Especially using `reduce`.

PR #2257, opened 2017-03-16
2256 2016 members interests

xybrnet:
Adds the json files for 2016 members' interests.


mhl:
@xybrnet Trying to import these on our staging site fails, in fact - I think because the `name` element of each `release` still says 2015:

```
django.db.utils.IntegrityError: duplicate key value violates unique constraint "interests_register_release_name_key"
DETAIL:  Key (name)=(Parliament Register of Members' Interests 2015) already exists.
```

xybrnet:
@mhl 
That commit message was vague - I missed that : )
I used a single quote (`) in it, which wasn't interpreted correctly.

I updated the year field, which fixed the duplicate key issue, and added the instructions to the README.

PR #2265, opened 2017-04-03
Disable Django template caching on Staging

zarino:
Templates were previously cached, no matter where the site was running, which meant restarting the `./manage.py runserver` process whenever you made a change to the HTML templates on a staging / dev site.

This commit disables the caching if the Staging flag is set.

### DIFF
--- a/pombola/settings/base.py
+++ b/pombola/settings/base.py
@@ -23,6 +23,20 @@ else:
 # switch on all debug when staging
 DEBUG          = STAGING
 
+template_loaders = [
+    'django.template.loaders.app_directories.Loader',
+    'django.template.loaders.filesystem.Loader',
+    # 'django.template.loaders.eggs.Loader',
+]
+
+if not STAGING:
+    template_loaders = [
+        (
+            'django.template.loaders.cached.Loader',
+            template_loaders,
+        )
+    ]
+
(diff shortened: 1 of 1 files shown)


pombola/settings/base.py
```diff
@@ -23,6 +23,21 @@
...
+    ]
+else:
+    template_loaders = [
+        ('django.template.loaders.cached.Loader', [
+            'django.template.loaders.app_directories.Loader',
+            'django.template.loaders.filesystem.Loader',
+            # 'django.template.loaders.eggs.Loader',
+        ]),
+    ]
+
```
dracos: I'd possibly rewrite this as:
```
template_loaders = [
    'django.template.loaders.app_directories.Loader',
    'django.template.loaders.filesystem.Loader',
    # 'django.template.loaders.eggs.Loader',
]
if not STAGING:
    template_loaders = [('django.template.loaders.cached.Loader', template_loaders)]
```
…so that if you change the base loaders in future, you only have to change it in one location.

zarino: Good idea!

mhl:
Looks good :+1: I've done some minor rewriting of the commit message (capitalizing STAGING where I think the config variable was intended)

PR #2272, opened 2017-04-20
2271 new attendance api

xybrnet:
As person's position can change at any time during a year, and attendance for ministers and MPs are reported differently, ministerial status should be checked at the date of each meeting.

A new API endpoint is used for data on the MP attendance page.

### DIFF
--- a/pombola/south_africa/views/attendance.py
+++ b/pombola/south_africa/views/attendance.py
@@ -230,14 +230,12 @@ class SAMpAttendanceView(TemplateView):
 
                         aggregate_total += total
                         aggregate_present += present
-
                         present_perc = self.calculate_abs_percenatge(present, total)
                         arrive_late_perc = self.calculate_abs_percenatge(arrive_late, total)
                         depart_early_perc = self.calculate_abs_percenatge(depart_early, total)
-
                         context['attendance_data'].append({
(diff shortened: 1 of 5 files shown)


pombola/south_africa/views/attendance.py
```diff
@@ -50,74 +51,113 @@ def just_path_of_pa_url(self, ma):
...
+                    if attendance_as_minister:
+                        minister_attendance.append({'member': record['member'], 'meetings': attendance_as_minister})
+                    if attendance_as_mp:
+                        mp_attendance.append({'member': record['member'], 'meetings': attendance_as_mp})
+                    minister_slugs.remove(slug)
+                else:
+                    # Member wasn't a minister during the year. All attendance as MP.
+                    mp_attendance.append(record)
+
+        if ctx_pos== 'ministers':
```
mhl: There's a missing space before `==` here.

xybrnet:
I've made the changes requested, thank you for them.

Darn. I'm not sure whether or not we should improve the logic if you're concerned. It is a bit urgent, and it would be good for them to view it sooner than later to check the accuracy of the results, but I would also prefer it to be sound.

It have done checks on a sample, and those have been accurate.

PR #2273, opened 2017-05-04
Swap Appearances and Attendance tabs

longhotsummer:
This is by urgent request from PMG. Attendance is far more of a hot topic in SA than appearances in committees, so they'd like it to come first. They have an article going out on Monday for which they'd like these changes.

They would also like to rename "Appearances" to "Comments and Questions", but that seemed like a larger and more controversial change (will the subheadings on that tab then change?), so I would like to get the swap in first.


longhotsummer:
Please note I don't have Pombola setup on my machine to check this, I'm making it directly in GitHub because of the urgency.

mhl:
Sorry, @longhotsummer I totally missed this PR :( I've merged and deployed it now.

Issue #2281, opened 2017-06-19
admins not getting notifications for new comments using the Facebook widget


mhl:
This was reported by @xybrnet:

> We've received an enquiry from [...] at People's Assembly, where she said that she's not receiving notifications for comments made on the PA site.

This might be to do with the Facebook app that we're using to record comments not being marked as "live", but this needs more investigation.

chrismytton:
There don't seem to be any comments appearing in the [comment moderation tool](https://developers.facebook.com/tools/comments/1619725741628189/pending/descending/). After a bit of digging it seems that this is because we need to add a `<meta property="fb:app_id" …>` tag, which is documented in the [comment moderation setup instructions](https://developers.facebook.com/docs/plugins/comments#moderation-setup-instructions).

chrismytton:
Additionally, [this Stack Overflow post](https://stackoverflow.com/a/34252960) suggests that once we've got the moderation tool working we'll need to go into the comment moderation tool settings and add People's Assembly as a "moderator" (and possibly remove them as an admin on the app).

chrismytton:
Also as @mhl notes in the original comment text, the app wasn't yet live. I've now toggled the switch in the app settings to make it live.

chrismytton:
I added the `fb:app_id` meta tag in https://github.com/mysociety/pombola/pull/2283 but it's still not being output on blog posts.

I did a bit more digging and found that [`pombola/south_africa/templates/info/blog_post.html#L17`](https://github.com/mysociety/pombola/blob/4d05dc88f3be6acfddc49b18cf4ccff907720557/pombola/south_africa/templates/info/blog_post.html#L17) overrides the open graph block and includes [this template](https://github.com/mysociety/django-info-pages/blob/2d6159090a61862ff5a6459c61678e68b9c66fe5/info/templates/info/_blog_open_graph.html) from django-info-pages.

So I think the best thing to do is to move the `fb:app_id` meta tag outside the `open_graph` template block so that it is _always_ in the HTML output.

chrismytton:
The `fb:app_id` meta tag is appearing on every page now https://github.com/mysociety/pombola/pull/2285 has been merged.

[...]

PR #2283, opened 2017-06-29
Output fb:app_id meta tag on all pages

chrismytton:
In order for the [Facebook comments moderation tool](https://developers.facebook.com/tools/comments/1619725741628189/pending/descending/) to work correctly this meta tag needs to be on all pages that contain the comment widget, more details [in the documentation](https://developers.facebook.com/docs/plugins/comments#moderation-setup-instructions). To future proof this I'm putting it into the base template, so that it gets included on all pages on the site.

I haven't got a good way of testing whether this actually makes the moderation tool work other than putting it live and checking, but suggestions welcome!

Part of https://github.com/mysociety/pombola/issues/2281

### DIFF
--- a/pombola/core/templates/core/object_base.html
+++ b/pombola/core/templates/core/object_base.html
@@ -1,12 +1,6 @@
 {% extends 'base.html' %}
 {% load thumbnail %}
 
-{% block extra_head_meta %}
-{% if settings.FACEBOOK_APP_ID %}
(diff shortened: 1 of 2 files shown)


chrismytton:
> could you link to the Facebook documentation (as you've done in the pull request message) in the commit message?

Ah yep, good idea! I've force pushed an amended commit linking to the relevant docs.

PR #2295, opened 2017-08-14
[ZA] Add redirect for /attendance to /mp-attendance

### LINKED ISSUE #2289: Redirect from https://www.pa.org.za/attendance TO https://www.pa.org.za/mp-attendance
As this link now shows attendance not only MPs but Ministers too, could we include a redirect from https://www.pa.org.za/attendance

Reason: It is confusing to reference the link https://www.pa.org.za/mp-attendance/ in a media article when we are talking about ministers

jacksonj04:
Closes #2289.

### DIFF
--- a/pombola/south_africa/urls.py
+++ b/pombola/south_africa/urls.py
@@ -159,6 +159,11 @@ urlpatterns += (
             url='/mp-attendance/',
             permanent=False)
     ),
+    url(r'^attendance/?$',
+        RedirectView.as_view(
+            url='/mp-attendance/',
+            permanent=False)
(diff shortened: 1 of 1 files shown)


pombola/south_africa/urls.py
```diff
@@ -159,6 +159,11 @@
             url='/mp-attendance/',
             permanent=False)
     ),
+    url(r'^attendance/$',
```
mhl: We're not using APPEND_SLASH on this project, so it's best to make the trailing slash optional, I think.

PR #2298, opened 2017-08-15
[ZA] Add tracking of usage of MP search box

### LINKED ISSUE #2259: ZA: add GA tracking of whether people use the filter box on the MP profiles page
As suggested by @zarino we should really find out whether people are using this box:

screenshot from 2017-03-24 18-27-25

on http://www.pa.org.za/position/member/parliament/ by tracking input events in Google Analytics

jacksonj04:
Closes #2259

### DIFF
--- a/pombola/south_africa/static/js/mp-profiles-live-filter.js
+++ b/pombola/south_africa/static/js/mp-profiles-live-filter.js
@@ -40,6 +40,7 @@ var showFilterResults = function showFilterResults(){
 };
 
 var filterProfiles = function filterProfiles(searchTerm, filterer){
+
     if(searchTerm === ''){
         return hideFilterResults();
     }
@@ -86,6 +87,14 @@ $(function(){
             $('.js-mp-profiles-live-filter').val(),
             filterer
         );
+    });
+
+    $(document).one('js-mp-profiles-live-filter:updated', function(){
+        window.analytics.trackEvent({
+          eventCategory: 'mp-name-search',
+          eventAction: 'search',
+          'transport': 'beacon'
+        });
     });
 
     // If device is short, and search input is in bottom half of the screen


pombola/south_africa/static/js/mp-profiles-live-filter.js
```diff
@@ -39,7 +39,19 @@ var showFilterResults = function showFilterResults(){
     return $results;
 };
 
+var analyticsEventSent = false;
+
 var filterProfiles = function filterProfiles(searchTerm, filterer){
+
+    if (!analyticsEventSent) {
+        ga('send', 'event', {
```
zarino: Why not use `window.analytics.trackEvent` from `js/analytics.js` (which gets concatenated into `base.js` by `pombola/base.py`)?

```javascript
window.analytics.trackEvent({
  eventCategory: 'mp-name-search',
  eventAction: 'search'
});
```

It’s what we used in the homepage A/B test, so I think

pombola/south_africa/static/js/mp-profiles-live-filter.js
```diff
@@ -39,7 +39,19 @@ var showFilterResults = function showFilterResults(){
     return $results;
 };
 
+var analyticsEventSent = false;
+
 var filterProfiles = function filterProfiles(searchTerm, filterer){
+
+    if (!analyticsEventSent) {
```
zarino: Maybe this is crazy, but you could achieve this "fire once" functionality without the global `analyticsEventSent` variable by hooking a `$.one()` listener onto the `js-mp-profiles-live-filter:updated` event that `filterProfiles` fires once it’s finished filtering the list. Off the top of my head, I think this would look like:

```javascript
$('body').one('js-mp-profiles-live-filter:updated', function(){ … });
```

jacksonj04: I had no idea that `.one()` was even a thing, good shout.

jacksonj04:
Scrutinizer failure due to long-running tests using all our Travis capacity, and coverage not arriving in time to make report. Merging anyway.

Issue #2303, opened 2017-09-06
[KE] Update Mzalendo homepage now 2017 election is over


zarino:
- [x] Remove "Election 2017" section
- [ ] In its place:
   - [x] Contact us offline button 21516. (This should be in red.)
   - [x] Our latest newsletter link. (We have about 3500 readers each week.)
   - [ ] Featured person of the week. (The spinner can have three people). The latest will always be the one featured in the newsletter. Kindly advise if this can be automated or we have to manually update each week.
   - [ ] Editorial/Blog - We can have our top two or three entries showing through each time.

Overall aim is to make the homepage feel "more alive, with recent Mzalendo material."

zarino:
https://github.com/mysociety/pombola/pull/2309 has implemented the first few points of this ticket.

![screen shot 2017-09-22 at 13 53 51](https://user-images.githubusercontent.com/739624/30745297-7a4d2024-9f9d-11e7-9321-57fe0a8e16c4.png)

Remaining tasks are:
* Featured person slider (not sure how we should implement this – @markl?)
* More than 2 blog posts?

zarino:
> Featured person slider (not sure how we should implement this – @mhl?)

Actually, I’ve just seen that the Kenyan homepage is already passed a `context['featured_persons']` array, containing a bunch of person objects (this is completely unused, and separate from the `context['featured_person']` used in the current sidebar widget). 

So I’m going to use that for now. Currently looking up a javascript slider widget that supports displaying, say, 3 "slides" (people) at a time, with infinite looping, and, if possible, Opera Mini support. The last one is gonna be hard.

zarino:
I’ve had clarification from Jessica that only one "featured person" needs to be visible at a time. So that simplifies the slider code somewhat (easier to make a single person responsive, than a three-column layout).

PR #2309, opened 2017-09-22
KE: Post-election homepage updates

zarino:
Part of #2303.

Here’s a screenshot, although the blog post previews are broken on my local machine, because CORS :-(

screen shot 2017-09-22 at 11 32 00


mhl:
Thanks, @zarino :+1: (that feeds.js error is a good catch)

PR #2310, opened 2017-09-22
[KE] Add script to import 2017 Election winners

jacksonj04:
NB: YNR data has been slightly massaged to correct minor issues like place punctuation.


pombola/kenya/management/commands/kenya_import_2017_election_results.py
```diff
@@ -0,0 +1,295 @@
...
+                        area = Place(
+                            slug=area_slug,
+                            name=row['post_label'],
+                            kind=import_type['area_type']
+                        )
+
+                    # If the area has a term, set that
+                    if import_type['area_has_term']:
+                        # This is a tuple from get_or_create, hence the [0]
+                        area.parliamentary_session = PARLIAMENTARY_SESSION[0]
```
mhl: I think it's probably less surprising to do the `[0]` dereferencing at the point where `PARLIAMENTARY_SESSION` is defined, then you don't even need the comment :)

mhl:
@jacksonj04 One other thing: what are you planning to do about ending the current positions?

Issue #2326, opened 2017-11-13
Allow sending messages via WriteInPublic from Pombola


chrismytton:
As part of https://github.com/mysociety/pombola/issues/1682 we want to be able to send messages via WriteInPublic from within a Pombola based site. This will mean using the WriteInPublic API to send messages, which is documented [in the integration guide](https://github.com/mysociety/write-it/blob/master/INTEGRATION_GUIDE.md#sending-messages).

This is probably best done as a new Django app within Pombola that encapsulates the interactions with WriteInPublic.

Part of https://github.com/mysociety/pombola/issues/1682

chrismytton:
While writing https://github.com/mysociety/pombola/issues/2327 I realised that we're going to need to store the sent message in the local database as well as sending it to the WriteInPublic API, so that when we retrieve a reply we can match it up to an original message.

chrismytton:
I've opened https://github.com/mysociety/pombola/pull/2329 with an early-stage proof-of-concept of this working.

chrismytton:
This is now working for writing to MPs. Going to close this ticket and open more specific issues for the outstanding bits.

Issue #2301, opened 2017-08-23
[ZA] add a tab on a person pages to show any correspondence on WriteInPublic


mhl:
We are currently working on setting up a WriteInPublic site for South African MPs, and one of the requirements is to link to the correspondence on that site from the appropriate MP's page on PA.

I haven't looked into how we'd do this yet, but WriteInPublic does have an API layer, so I would hope we would be able to fetch the messages for a person from that. This will almost certainly involve making sure that that there is some common ID between the two systems as well.

mhl:
If someone has no correspondence on this tab, there should be a call-to-action to write to them.

chrismytton:
Done in #2346

PR #2329, opened 2017-11-15
[WIP] Send messages via writeinpublic

chrismytton:
### Outstanding tasks

- [x] Use the correct person ID, rather than a hard-coded one
- [x] Move the WriteInPublic URL, instance ID, username and API key into the config (currently all hard-coded)
- [x] Redirect to message page after creation
- [x] Display message replies on the messages page
- [x] Add message preview step
- [ ] Write some tests for the form and API client
- [x] Breadcrumbs on individual message page?? ("Write" link current 404s)
- [x] Embed list of messages _into_ representative profile page (as a tab)? Or ajax it in? #2301

## Notes to merger

- [x] https://github.com/mysociety/pombola/pull/2331 need to be un-blocked and merged before this pull request can be merged.
- [x] https://github.com/mysociety/write-it/pull/2 needs to be reviewed and merged in write-it

Fixes https://github.com/mysociety/pombola/issues/2326


JenMysoc:
PMG have asked if we could add in the following text:

at the end when it states: Success, your message has now been sent, there should be a: "Return to Write to my MP" link rather than "Correct this Page" link at the bottom of page.

chrismytton:
This was split up into separate pull requests (the "merged" PRs referenced above), so the code from this is now on `master`.

Issue #2330, opened 2017-11-20
[ZA] Sync people's EveryPolitician UUID to the local database


chrismytton:
As part of #1682 we need to know what each person's EveryPolitician UUID is so that we can provide it when creating and retrieving messages for that person.

chrismytton:
Pull request for this here: https://github.com/mysociety/pombola/pull/2331

However it's not ready to be merged yet as it's currently pointing to an everypolitician-data branch which is losing data, so we need to fix https://github.com/everypolitician/everypolitician-data/issues/42096 first before that can be merged.

PR #2331, opened 2017-11-24
Sync EveryPolitician UUID to Person model

chrismytton:
Add a management command add EveryPolitician UUIDs to the Person model.

### ~~Blocked waiting for EveryPolitician South Africa import to be fixed~~

~~This is currently using the `south-africa-update-peoples-assembly` branch of everypolitician-data, which has the required UUIDs but is also losing some data, so can't be merged in its current state. So before this can be merged we need to fix the problems described in https://github.com/everypolitician/everypolitician-data/issues/42096.~~

To work around this problem I've made the git ref that we use to fetch `countries.json` configurable. This means we can point to the branch that has the data we need for now, then when things are fixed upstream we can use a different ref to point to the fixed data.

I've also split the bit that adds a cronjob for this into a separate pull request https://github.com/mysociety/pombola/pull/2342 so we can hold off merging that until we've got the data in EveryPolitician up-to-date, but we can still merge this in the meantime.

Fixes https://github.com/mysociety/pombola/issues/2330

### DIFF
--- a/pombola/south_africa/management/commands/south_africa_sync_everypolitician_uuid.py
+++ b/pombola/south_africa/management/commands/south_africa_sync_everypolitician_uuid.py
@@ -1,0 +1,46 @@
+from everypolitician import EveryPolitician
+
+from django.core.management.base import BaseCommand
+
+from pombola.core.models import Person
+
+
+class Command(BaseCommand):
+    help = "Sync EveryPolitician UUID to Person's identifiers array"
+
+    def add_arguments(self, parser):
+        parser.add_argument('everypolitician_countries_json_git_ref',
+                            default='master', nargs='?',
+                            help="A git ref from the everypolitician-data repo")
+
+    def handle(self, **options):
+        verbose_level = options['verbosity']
+
+        url_template = ('https://cdn.rawgit.com/everypolitician/everypolitician-data/'
+                        '{git_ref}/countries.json')
+
+        url = url_template.format(git_ref=options['everypolitician_countries_json_git_ref'])
+
+        ep = EveryPolitician(countries_json_url=url)
+        south_africa_assembly = ep.country('South-Africa').legislature('Assembly').popolo()
+
+        id_lookup = {}
+        for popolo_person in south_africa_assembly.persons:
+            id_lookup[popolo_person.identifier_value('peoples_assembly')] = popolo_person.id
+
+        error_msg = u"No EveryPolitician UUID found for {0.id} {0.name} https://www.pa.org.za/person/{0.slug}/\n"
+        for person in Person.objects.filter(hidden=False):
+            uuid = id_lookup.get(str(person.id))
+            if uuid is None:
+                verbose_level > 1 and self.stderr.write(error_msg.format(person))
+                continue
+            identifier, created = person.identifiers.get_or_create(
+                scheme='everypolitician',
+                identifier=uuid,
+            )
+            if verbose_level > 0:
+                if created:
+                    msg = u"Created new identifier for {name}: {identifier}"
+                else:
+                    msg = u"Existing identifier found for {name}: {identifier}"
+                self.stdout.write(msg.format(name=person.name, identifier=identifier.identifier))
--- a/requirements.txt
+++ b/requirements.txt
@@ -173,3 +173,7 @@ gunicorn==19.4.5
 whitenoise==2.0.6
 
 cffi
+
+# everypolitician packages
+everypolitician==0.0.13
+everypolitician-popolo==0.0.11


chrismytton:
I think it would be good to get this merged ASAP rather than waiting for the EP/DC problems to be resolved. The best way to do this is probably to make the SHA configurable where there's currently a `TODO` comment in this change:

https://github.com/mysociety/pombola/blob/4ea415a847ffddeaa2c4609e6f4aa8f67fc53293/pombola/core/management/commands/core_sync_everypolitician_uuid.py#L18-L20

Then this script will be usable with whatever EveryPolitician data we want to use.

chrismytton:
Have extracted the crontab bits out to https://github.com/mysociety/pombola/pull/2342 so that the code for the management command can be merged now, and then we can add the cronjob once we're happy with the EP data.

pombola/core/management/commands/core_sync_everypolitician_uuid.py
```diff
@@ -0,0 +1,54 @@
...
+                            help="A git ref from the everypolitician-data repo")
+
+    def handle(self, **options):
+        verbose_level = options['verbosity']
+
+        # FIXME: This SHA is currently pointing at a branch which has the
+        # `peoples_assembly` identifier, but this isn't on master because
+        # South Africa is broken in EveryPolitician at the moment.
+        #
+        # TODO: Make this SHA configurable.
```
mhl: I think these comments can be removed now :)

pombola/core/management/commands/core_sync_everypolitician_uuid.py
```diff
@@ -0,0 +1,54 @@
...
+                        '{git_ref}/countries.json')
+
+        url = url_template.format(git_ref=options['everypolitician_countries_json_git_ref'])
+
+        ep = EveryPolitician(countries_json_url=url)
+        south_africa_assembly = ep.country('South-Africa').legislature('Assembly').popolo()
+
+        id_lookup = {}
+        for popolo_person in south_africa_assembly.persons:
+            pa_ids = [i for i in popolo_person.identifiers if i['scheme'] == 'peoples_assembly']
```
mhl: I think you can use `popolo_person.identifier_value('peoples_assembly')` to simplify this: https://github.com/everypolitician/everypolitician-popolo-python/blob/master/popolo_data/base.py#L99 (this isn't documented, sorry)

pombola/core/management/commands/core_sync_everypolitician_uuid.py
```diff
@@ -0,0 +1,54 @@
...
+            if pa_ids == []:
+                continue
+            id_lookup[pa_ids[0]['identifier']] = popolo_person.id
+
+        error_msg = u"No EveryPolitician UUID found for {} {} https://www.pa.org.za/person/{}/\n"
```
mhl: This is partly just "fun with Python's `.format`" but, you could make this:
```
u"No EveryPolitician UUID found for {0.id} {0.name} https://www.pa.org.za/person/{0.slug}/\n"
```
... and then do `error_msg.format(person)`. It's a tiny thing, but it'd make changing the format string later less prone to error.

pombola/core/management/commands/core_sync_everypolitician_uuid.py
```diff
@@ -0,0 +1,54 @@
...
+        for person in Person.objects.filter(hidden=False):
+            uuid = id_lookup.get(str(person.id))
+            if uuid is None:
+                if verbose_level > 1:
```
mhl: I'd probably collapse these two conditions into a single condition - perhaps makes it slightly easier to see that there's only one branch here?

pombola/core/management/commands/core_sync_everypolitician_uuid.py
```diff
@@ -0,0 +1,54 @@
+from everypolitician import EveryPolitician
+
+from django.core.management.base import BaseCommand
+
+from pombola.core.models import Person
+
+
+class Command(BaseCommand):
+    help = 'Sync EveryPolitician UUID to Person.everypolitician_uuid field'
```
mhl: "Person.everypolitician_uuid field" - I suspect this is an old comment?

pombola/core/management/commands/core_sync_everypolitician_uuid.py
```diff
@@ -0,0 +1,54 @@
...
+                continue
+            identifier, created = person.identifiers.get_or_create(
+                scheme='everypolitician',
+                identifier=uuid,
+            )
+            if verbose_level > 0:
+                if created:
+                    msg = u"Created new identifier for {}: {}"
+                else:
+                    msg = u"Existing identifier found for {}: {}"
```
mhl: It's another very minor thing, but I'd probably make these use `{name}: {identifier}` and then `.format(name=person.name, identifier=identifier.identifier)

pombola/core/management/commands/core_sync_everypolitician_uuid.py
chrismytton: :+1:

pombola/core/management/commands/core_sync_everypolitician_uuid.py
chrismytton: Oops, yep!

pombola/core/management/commands/core_sync_everypolitician_uuid.py
chrismytton: Ah didn't know about that, thanks, have changed it to use that.

pombola/core/management/commands/core_sync_everypolitician_uuid.py
chrismytton: 😮 Wow, didn't know `.format` could do that!! You're quite right, that's less error prone. Have switched to using that syntax.

pombola/core/management/commands/core_sync_everypolitician_uuid.py
chrismytton: The `continue` on the line below is actually in the top level `if` statement, then the logging is part of the inner `if` statement. I've changed things around in https://github.com/mysociety/pombola/pull/2331/commits/663e8dc2af62e5c38f490debc34ed0c423943db7 so that it's all at a single level of `if` statement to make things hopefully a bit clearer.

pombola/core/management/commands/core_sync_everypolitician_uuid.py
chrismytton: :+1:

chrismytton:
> The only other thing that occurs to me is that it might helpful if the argument was optional rather than required, and have it default to master.

Ah yep, very sensible suggestion! Have done that in https://github.com/mysociety/pombola/pull/2331/commits/e7e2e96828a4f81e48fd3781a058952302be43af.

chrismytton:
@mhl I've addressed you're very helpful comments, would you mind re-reviewing this please?

mhl:
@chrismytton Looks good :+1: - just one last thing; since this is currently South Africa specific, could you make the filename `pombola/core/management/commands/core_sync_everypolitician_uuid.py` instead?

chrismytton:
> @chrismytton Looks good 👍 - just one last thing; since this is currently South Africa specific, could you make the filename pombola/core/management/commands/core_sync_everypolitician_uuid.py instead?

I'm guessing that should have been `pombola/south_africa/management/commands/south_africa_sync_everypolitician_uuid.py`? If so I've done that in 2e63d6637a5b399ddc61ffaacffa84c505bd3ba5.

mhl:
Yep, that's what I meant, sorry :)

PR #2332, opened 2017-12-01
[ZA] Select correct person page tab on pageload

zarino:
While working on #2329, I noticed that the selected tab on a person page wasn’t persisted across page refreshes, despite the tab ID being in the URL hash fragment.

This fixes that, and lets you, for example, share a link directly to a Representative’s "Positions", or "Appearances" … or (once #2329 is merged) "Messages".


chrismytton:
This has now been deployed to https://www.pa.org.za/.

PR #2340, opened 2018-01-17
Update Dokeza wording

JenMysoc:
changing wording on Dokeza button as per Jessica's instructions


zarino:
Here’s how it looks 😭 

screen shot 2018-01-17 at 10 36 10

JenMysoc:
sigh :( we don't have time to fix that now so we'll just have to live with it until we have a little more mzalendo time

zarino:
In the grand scheme of things, I’ve seen much worse. So as a temporary solution, we might as well deploy it.

PR #2342, opened 2018-01-31
[South Africa] Add cron job to sync EP UUIDs

chrismytton:
Extracted from https://github.com/mysociety/pombola/pull/2331

This adds a cron job that runs the `core_sync_everypolitician_uuid` management command that was added in #2331 once a day on staging and production.

## Notes to merger

- [x] Don't merge this until #2331 has been merged, as it uses code from that pull request.

### DIFF
--- a/conf/crontab.ugly
+++ b/conf/crontab.ugly
@@ -68,6 +68,9 @@ MAILTO=cron-!!(*= $site *)!!@mysociety.org
 # NOTE - this should probably be removed once the live site is launched
 13 0 * * * !!(*= $user *)!! output-on-error update_za_hansard.bash
 
+# Sync EveryPolitician UUIDs to local DB
+0 10 * * * !!(*= $user *)!! output-on-error run_management_command south_africa_sync_everypolitician_uuid
+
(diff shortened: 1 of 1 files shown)


conf/crontab.ugly
```diff
@@ -77,5 +80,8 @@ MAILTO=cron-!!(*= $site *)!!@mysociety.org
 13 2 * * * !!(*= $user *)!! run_management_command core_export_to_popolo_json /data/vhost/!!(*= $vhost *)!!/media_root/popolo_json/ http://www.pa.org.za
 30 3 * * * !!(*= $user *)!! run_management_command core_export_to_popolo_json --pombola /data/vhost/!!(*= $vhost *)!!/media_root/popolo_json/ http://www.pa.org.za
 
+# Sync EveryPolitician UUIDs to local DB
+30 10 * * * !!(*= $user *)!! output-on-error run_management_command core_sync_everypolitician_uuid
```
mhl: I think the management command has a different name now - shouldn't it be: `south_africa_sync_everypolitician_uuid`?

chrismytton: Oh yes, of course! Good catch! 😟

Issue #1682, opened 2015-06-22
Add WriteInPublic to pa.org.za


paullenz:

mhl:
Probably the simplest thing to do is to create a new nightly Popolo export that only includes the politicians you'd want to be able to write to, then have a new PopIt instance that imports that, as we do with the current PopIt mirrors. The WriteInPublic could be based on that.

Then the question is whether this needs to appear to be embedded in the site or it's OK for it to look like a separate service. (The former would need more work.)

The same comments apply to https://github.com/mysociety/pombola/issues/1480

paullenz:
For ease, I would suggest a separate service that is styled to look like
the main pa site and linked to from it

JenMysoc:
From PMG: At this stage in our democracy, it would be a step too far to include WARD COUNCILLORS. At most, National Assembly MPs and possibly NCOP and MPLs [should be included in this tool]

JenMysoc:
From PMG: During our first 2017 Skype conversation with you, we talked about adding the Write to Them feature this year to each MP's profile. Can we assign this to the next sprint? Following the latest political dramas in SA, there has been an increased call for people’s  ability to be able to reach out to their MPs. During the Skype conversation we talked about the being able to moderated  the comments made to the MP.

This is now a high priority for them and if we could discuss when we might be able to look at this it would be very useful! It would be great to capitalise on the increased interest in contacting MPs :)

JenMysoc:
To do this I think we need to:

- [ ] Set up an instance of WiP including the ZA parliament and their emails
- [ ] Skin the instance of WiP with something so it looks like pa.org.za
- [ ] Create an admin account or allow multiple admin accounts so that people from PMG can moderate messages and send them on
- [ ] Import messages from Mp message pages back into PA

tmtmtmtm:
Assuming the WiP is going to be backed by EP, then we also need to fix some of the issues that are making it very difficult to keep EP in sync with PA, most importantly an equivalent of what we did for Mzalendo in https://github.com/mysociety/pombola/pull/2172

JenMysoc:
This has been paired down and now PMG want the following;

A WriteIt API instance that allows people to send emails to the central email address for a committee, some of these emails should either be pre-formatted or have simple questions in a form like element to enable ease of creating a message for people who aren't sure what to say. The form should be on the homepage or similar.

[...]

PR #2344, opened 2018-02-06
Add WriteInPublic API client

chrismytton:
This is a very simple wrapper around the WriteInPublic API. Currently it only implements the features that we need for integrating it into Pombola.

I've added some tests for the basic functionality, but in the interests of time I've skipped writing tests for the `.people()` and `.answers()` methods, so will need to revisit that at some point in the future.

Extracted from #2329 
Part of #1682

### DIFF
--- a/pombola/writeinpublic/client.py
+++ b/pombola/writeinpublic/client.py
@@ -1,0 +1,99 @@
+import requests
+
+from django.utils.dateparse import parse_datetime
+
+from pombola.core.models import Person
+
+
+class PersonMixin(object):
+    def parse_everypolitician_uuid(self, resource_uri):
+        return resource_uri.split('#')[-1].split('-', 1)[-1]
+
+
+class Message(PersonMixin, object):
+    def __init__(self, params):
+        self.id = params['id']
+        self.author_name = params['author_name']
+        self.subject = params['subject']
+        self.content = params['content']
+        self.created_at = parse_datetime(params['created'])
+        self._params = params
+
+    def people(self):
+        return Person.objects.filter(
+            identifiers__scheme='everypolitician',
+            identifiers__identifier__in=self._recipient_ids(),
+        )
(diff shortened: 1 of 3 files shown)


pombola/writeinpublic/client.py
```diff
@@ -0,0 +1,95 @@
...
+            'author_email': author_email,
+            'subject': subject,
+            'content': content,
+            'writeitinstance': "/api/v1/instance/{}/".format(self.instance_id),
+            'persons': persons,
+        }
+        try:
+            return requests.post(url, json=payload, params=params)
+        except requests.exceptions.RequestException as err:
+            raise self.WriteInPublicException(err)
```
mhl: The argument to this exceptions constructor is expected to be a string, I think, so perhaps this should be:
```
raise self.WriteInPublicException(unicode(err))
```

pombola/writeinpublic/client.py
```diff
@@ -0,0 +1,95 @@
...
+
+    def get_messages(self, person_popolo_uri):
+        url = '{url}/api/v1/instance/{instance_id}/messages/'.format(url=self.url, instance_id=self.instance_id)
+        params = {
+            'format': 'json',
+            'username': self.username,
+            'api_key': self.api_key,
+            'person__popolo_uri': person_popolo_uri,
+        }
+        response = requests.get(url, params=params)
```
mhl: I think that all calls to `requests.get` or `requests.post` should be followed by `response.raise_for_status()` afterwards, or you won't get an exception for 4xx and 5xx responses, and will get a less helpful error later...

chrismytton: Ah good shout, I didn't know about that method, added that in de1f549.

chrismytton:
@mhl I've made those changes now, so I think this is ready for another review.

PR #2345, opened 2018-02-08
WriteInPublic views

chrismytton:
This adds the three main views for the WriteInPublic integration:

- Create new message wizard
- Display an individual message
- Display a list of messages

**Note**: This doesn't include the `urls.py` entries to actually make these appear in the web interface, I've left these off so we can get this reviewed and merged without actually "launching" the WriteInPublic integration quite yet. The `urls.py` bits are in https://github.com/mysociety/pombola/pull/2347.

I realise that this is currently lacking tests 😞 I would like to remedy that in the near future, but in the interests of getting this launched by Friday I think it's better to get *something* in place now, then w can iterate on it and add tests etc.

Extracted from #2329 
Part of #2326


pombola/writeinpublic/forms.py
```diff
@@ -0,0 +1,28 @@
...
+
+class RecipientForm(forms.Form):
+    persons = ModelMultipleChoiceField(queryset=None)
+
+    def __init__(self, *args, **kwargs):
+        super(RecipientForm, self).__init__(*args, **kwargs)
+        self.fields['persons'].queryset = Person.objects.filter(
+            identifiers__scheme='everypolitician',
+            identifiers__identifier__isnull=False,
+        )
```
mhl: I think I'd add a comment here explaining why you're setting the queryset dynamically. (I think I see why, but I had to think about it quite hard!)

chrismytton:
> I think my bigger concern here is how we configure which WriteInPublic instance to use, and at the moment there's just one possible that's specified in general.yml. I think I've mentioned this before, but I think it's very likely we'll want to support using multiple WriteInPublic instances from Django. (Won't the committees be a second instance, for example?)
> 
> I haven't thought this through completely, but what came to my mind is that there could be a model in this new app that represents a particulary combination of WriteInPublic URL and instance ID, with username and API key. Then the ID of that model could be part of the URLs defined by this app, so you have separate URLs for using each instance.
> 
> I'm a bit conflicted about suggesting this because I know we want to get this change out of the door as soon as possible, but if we want to move to this way of supporting multiple write-it instances shortly, then it'll be a pain having to redirect the old URLs that just assume a single instance, for example.
> 
> What do you think?

@mhl and I discussed this on a call, we decided that in the interests of getting this merged we should leave it as it is for now, and make working with multiple instances something to worry about next week. @mhl has kindly offered to write up an issue about this so it doesn't get forgotten about.

pombola/writeinpublic/forms.py
chrismytton: :+1: I've added a comment about this in https://github.com/mysociety/pombola/pull/2345/commits/04950200b8af7dbf970163f88e8b560b77b0043e.

mhl:
One other thing - I've noted what you've said about the lack of tests, and that's fine for this week, but it'd be good to come back to that too.

Issue #2365, opened 2018-02-15
Move WriteInPublic instance config to database


chrismytton:
As suggested in https://github.com/mysociety/pombola/pull/2345#pullrequestreview-95158041 we should move the WriteInPublic configuration into a model instance so we can have multiple WriteInPublic instances running on the same Pombola instance.

To do this without disrupting service I think we'll need to break it into multiple steps:

1. Add a model to the `writeinpublic` app which holds a combination of WriteInPublic URL, instance ID, username and API key.
2. Deploy that change to production and create a new row in the database for the current WiP instance on pa.org.za
3. Change the views to use the new model instance instead of the `general.yml` values
4. Deploy that change to production and then remove the old config values

Once this is done it will allow us to have multiple WriteInPublic instances, which will unlock writing to committees #2328.

Part of https://github.com/mysociety/pombola/issues/2351

chrismytton:
Pull request for adding the model here: #2367, and for changing the views to use that here: #2368.

chrismytton:
Final pull request to remove the now unused settings here: #2374

PR #2368, opened 2018-02-15
Update WriteInPublic views to use configuration values

chrismytton:
Following on from #2367 this changes the view to get the configuration from the database, rather than the config file.

Follows on from #2367 
Second part of https://github.com/mysociety/pombola/issues/2365

## Notes to merger

Please do the following before merging

- [x] Once #2367 has been merged change the base branch to `master`
- [x] Deploy #2367 to the production and create a `Configuration` instance in the DB with the current production settings using the slug "south-africa-assembly".

### DIFF
--- a/pombola/south_africa/urls.py
+++ b/pombola/south_africa/urls.py
@@ -273,7 +273,8 @@ urlpatterns += (
     url(
         r'^person/(?P<person_slug>[-\w]+)/messages/$',
         WriteToRepresentativeMessages.as_view(),
+        kwargs={'configuration_slug': 'south-africa-assembly'},
         name='sa-person-write-all'
     ),
-    url(r'^write/', include('pombola.writeinpublic.urls')),
+    url(r'^write/', include('pombola.writeinpublic.urls'), kwargs={'configuration_slug': 'south-africa-assembly'}),
 )
--- a/pombola/writeinpublic/views.py
+++ b/pombola/writeinpublic/views.py
@@ -11,6 +11,7 @@ from pombola.core.models import Person
 
 from .forms import RecipientForm, DraftForm, PreviewForm
 from .client import WriteInPublic
+from .models import Configuration
 
 
 def person_everypolitician_uuid_required(function):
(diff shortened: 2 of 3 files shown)


pombola/writeinpublic/views.py
```diff
@@ -26,14 +27,15 @@ def wrap(request, *args, **kwargs):
 
 
 class WriteInPublicMixin(object):
-    def __init__(self, *args, **kwargs):
+    def get(self, *args, **kwargs):
```
chrismytton: This should be overriding the `dispatch` method, not the `get` method, otherwise it won't work with form POSTs.

pombola/south_africa/urls.py
```diff
@@ -273,7 +273,8 @@
     url(
         r'^person/(?P<person_slug>[-\w]+)/messages/$',
         WriteToRepresentativeMessages.as_view(),
+        {'configuration_slug': 'south-africa-assembly'},
         name='sa-person-write-all'
     ),
-    url(r'^write/', include('pombola.writeinpublic.urls')),
+    url(r'^write/', include('pombola.writeinpublic.urls'), {'configuration_slug': 'south-africa-assembly'}),
```
chrismytton: Feedback: Change the final parameter to `kwargs=` if that is equivalent to make it clearer what's going on.

pombola/writeinpublic/views.py
chrismytton: It turns out that `self.client` is never called from a POST request. When the POST is done to the "preview" step the wizard then redirects to the "done" step, so the browser does a GET request to that step, which is where `self.client` is used.

I've decided to change this to `dispatch` _anyway_, because it seems like a more robust approach for the future.

chrismytton:
I've added a (slightly convoluted) test for the wizard functionality. This tests all of the steps of the wizard to ensure the message is sent at the end of the process.

chrismytton:
Just waiting for Travis to finish then I'll request another review of this.

pombola/south_africa/urls.py
```diff
@@ -273,7 +273,8 @@
     url(
         r'^person/(?P<person_slug>[-\w]+)/messages/$',
         WriteToRepresentativeMessages.as_view(),
+        {'configuration_slug': 'south-africa-assembly'},
```
mhl: You could add the `kwargs=` here as well

chrismytton:
Ah I didn't know about django-webtest, thanks for the pointer @mhl! I'll definitely remember that for next time 👍

PR #2374, opened 2018-02-20
Remove WriteInPublic configuration

chrismytton:
As of #2368 these settings are stored in the database. Removing them from the settings files to reduce unnecessary clutter.

## Notes to deployer

- You will need to remove these settings from the `vhosts/pombola_conf_general.yml.ugly` file in `/data/servers` when deploying this to the mySociety servers.

Fixes https://github.com/mysociety/pombola/issues/2365


chrismytton:
This has now been deployed to production.

PR #2380, opened 2018-02-22
[ZA] Write to a committee

### LINKED ISSUE #1682: Add WriteInPublic to pa.org.za

chrismytton:
This adds the ability to write to a committee alongside the existing Write to an MP functionality.

Fixes #1682

### DIFF
--- a/pombola/writeinpublic/views.py
+++ b/pombola/writeinpublic/views.py
@@ -119,6 +119,8 @@ class WriteInPublicMixin(object):
         configuration_slug = kwargs['configuration_slug']
         configuration = Configuration.objects.get(slug=configuration_slug)
 
+        self.app_name = kwargs.get('app_name')
+
         # FIXME: It would be nice if we didn't hardcode the configuration_slug
         # values here.
         if configuration_slug == 'south-africa-assembly':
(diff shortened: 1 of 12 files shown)


pombola/writeinpublic/views.py
```diff
@@ -55,11 +55,41 @@ def object_ids(self, objects):
...
+        configuration = Configuration.objects.get(slug=configuration_slug)
+
+        # FIXME: It would be nice if we didn't hardcode the configuration_slug
+        # values here.
+        if configuration_slug == 'south-africa-assembly':
+            self.adapter = PersonAdapter()
+        elif configuration_slug == 'south-africa-committees':
+            self.adapter = CommitteeAdapter()
+        else:
+            raise Exception, "Unknown configuration_slug: {}".format(configuration_slug)
```
mhl: There's a minor PEP 8 thing here - it recommends using `raise Exception('message')` instead of `raise Exception, 'message'`. However, I think I'd write this differently anyway, more like:
```python
self.adapter = {
    'south-africa-assembly': PersonAdapter,
    'south-africa-committees': CommitteeAdapter,
}[configuration_slug]()
```

Issue #2324, opened 2017-11-08
ZA (WriteInPublic): create committee secretaries Popolo JSON file suitable for WIP


mhl:
One of the instances that PMG want is just the secretaries of committees. They've supplied us with the email addresses of these scretaries in a word document, but I've turned that into a spreadsheet here:

https://docs.google.com/spreadsheets/d/1-JNEZcHhO99UuUQNFPjKnuE-lhv6Ddj6yiapfaBC5r0/edit#gid=0

We will need to generate a Popolo JSON file with these details suitable for presenting them as options to write to in WriteInPublic

chrismytton:
I've been investigating https://github.com/mysociety/pombola/issues/1216 which is about cleaning up the committees in Pombola South Africa. From my investigations it seems that the `OrganisationKind` that we'll need to use for getting the list of committees for adding the email addresses and eventually for writing to is name == "National Assembly Committee".

chrismytton:
This is now done. Committee Popolo JSON available here: https://www.pa.org.za/api/committees/popolo.json

PR #2381, opened 2018-02-23
[ZA] Add API endpoint to serve committee Popolo JSON

chrismytton:
This adds an API endpoint which serves up Popolo JSON suitable for feeding into WriteInPublic. This means that any changes to committee email addresses will automatically be reflected in WiP when the datasource is synced.

## Notes to reviewer

Currently there aren't any committees with an email address in the production pa.org.za deploy. I added email addresses to a couple of committees locally to test this. Below is an example of the JSON that this call produces, note that the empty `contact_details` array is needed for WriteInPublic, otherwise an error is produced on import.

```json
{
  "persons": [
    {
      "email": "test@example.com",
      "contact_details": [],
      "id": 1410,
      "name": "Ad Hoc Committee on Funding of Political Parties"
    },
    {
      "email": "committee@example.org",
      "contact_details": [],
      "id": 141,
      "name": "Ad Hoc Committee on Code of Judicial Conduct on Disclosure of Interests (NA)"
    }
  ]
}
```

Part of https://github.com/mysociety/pombola/issues/2324

### DIFF
--- a/pombola/south_africa/views/api.py
+++ b/pombola/south_africa/views/api.py
@@ -1,0 +1,28 @@
+from django.http import JsonResponse
+from django.views.generic import ListView
+
+from pombola.core.models import Organisation
+
+
+# Output Popolo JSON suitable for WriteInPublic for any committees that have an
+# email address.
+class CommitteesPopoloJson(ListView):
+    queryset = Organisation.objects.filter(
+        kind__name='National Assembly Committees',
+        contacts__kind__slug='email'
+    )
+
+    def render_to_response(self, context, **response_kwargs):
+        return JsonResponse(
+            {
+                'persons': [
+                    {
+                        'id': committee.id,
+                        'name': committee.name,
+                        'email': committee.contacts.filter(kind__slug='email')[0].value,
+                        'contact_details': []
+                    }
+                    for committee in context['object_list']
+                ]
+            }
+        )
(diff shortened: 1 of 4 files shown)


pombola/south_africa/views/api.py
```diff
@@ -0,0 +1,21 @@
...
+        contacts__kind__slug='email'
+    )
+    popolo = []
+    for committee in committees:
+        popolo.append({
+            'id': committee.id,
+            'name': committee.name,
+            'email': committee.contacts.filter(kind__slug='email')[0].value,
+            'contact_details': []
+        })
```
mhl: I think patterns like this (create a list and append to it in a loop) are less error prone to write as list comprehensions

pombola/south_africa/views/api.py
```diff
@@ -0,0 +1,21 @@
+from django.http import JsonResponse
+
+from pombola.core.models import Organisation, ContactKind
+
+
+# Output Popolo JSON suitable for WriteInPublic for any committees that have an
+# email address.
+def committees_popolo_json(request):
```
mhl: I'd slightly prefer this to be a class-based view than a function, for consistency with the rest of `pombola.south_africa.views` - there are function-based views in Pombola core, but generally, I think we're trying to move towards class-based views. (It doesn't really matter for a view as simple as this, but I think they help to make the code clearer in more complex cases.) There's a similar example to this in `id-mapping-store`, which might be a useful comparison: https://github.com/mysociety/id-mapping-store/blob/627a9b07f06122da7a80cfbf3bb75fab36b90ef4/id_mappings/views.py#L118-L134

chrismytton:
Discussed these changes on a call with @mhl and he gave this the green light.

Issue #2351, opened 2018-02-09
Support multiple WriteInPublic instances being used from Pombola


mhl:
With the current writeinpublic Django app in Pombola, you can only configure one instance to use, which is done in `general.yml-example`.

It would be nice instead if one could add new WriteInPublic instances in the Django admin, and be able to use different instances depending on which instance slug is used in the URL.

This probably means adding a model in `pombola/writeinpublic/models.py` a model that has fields for:
* The WriteInPublic instance's base URL
* The WriteInPublic owner's username
* The WriteInPublic API Key
* The WriteInPublic instance ID (and / or slug?)

chrismytton:
As of https://github.com/mysociety/pombola/pull/2384 this is now possible.

PR #2384, opened 2018-02-26
[WriteInPublic] Use URL namespaces to allow multiple app instances

chrismytton:
Namespacing the URLs allows us to "mount" multiple instances of the writeinpublic app in the URL config. This means we can have one instance for writing to MPs and another for writing to committees.

This adds namespacing support to the `writeinpublic` app without changing the way it works. This change doesn't add support for committees, that will be added in a separate pull request.

Part of https://github.com/mysociety/pombola/issues/2351

### DIFF
--- a/pombola/south_africa/templates/core/person_detail.html
+++ b/pombola/south_africa/templates/core/person_detail.html
@@ -31,7 +31,7 @@
     <div class="contact-actions">
       {% if person.everypolitician_uuid %}
         <div class="contact-actions__wip">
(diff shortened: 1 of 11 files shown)


chrismytton:
> Did you mean to include the commented out link to the start of the process in the committees namespace in this commit?

Yes, the link was previously there but commented out in HTML, which meant Django still tried to render it. Rather than update the already incorrect link I commented it out in at the Django template level and pointed it at the (yet to be implemented from the pull request's POV) committees WriteInPublic integration.

Perhaps I should have just removed the link for the time being, but it's going to be used really soon, so I think I'm OK with leaving it in :)

Issue #2328, opened 2017-11-13
[ZA] Show WriteInPublic messages sent to committees


chrismytton:
As part of https://github.com/mysociety/pombola/issues/1682 we want to add an index page listing committees, linking through to all WriteInPublic correspondence sent to them.

PR #2389, opened 2018-03-06
[WriteInPublic] Make view more generic

chrismytton:
Rather than hardcoding the `Person` class all over the place this instead uses an adapter approach, which means we can pass in different adapters to the view which can use different models to get the required information.

This will allow us to have two different WriteInPublic instances, one that uses `Person` models to write to people and another, which is yet to be added, which will allow you to write to a committee, which is represented by the `Organisation` model.

Part of https://github.com/mysociety/pombola/issues/2328

### DIFF
--- a/pombola/writeinpublic/tests.py
+++ b/pombola/writeinpublic/tests.py
@@ -1,5 +1,8 @@
+from datetime import datetime
+
 import requests_mock
 from nose.plugins.attrib import attr
+from mock import Mock
 
 from django.test import TestCase
 from django.core.urlresolvers import reverse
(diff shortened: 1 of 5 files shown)


pombola/writeinpublic/tests.py
```diff
@@ -14,7 +15,15 @@
 @requests_mock.Mocker()
 class ClientTest(TestCase):
     def setUp(self):
-        self.writeinpublic = client.WriteInPublic('https://example.com', 'test', '123', '42', 'https://example.net/p.json#person-{}')
+        adapter = Mock()
```
mhl: Since you're creating this mock adapter before each test, could you store this as `self.adapter` and then make assertions about the calls you expect it to receive in each test below?

chrismytton:
> This is one of those pull requests where I'm interested to see where it's going next (for example, what the OrganisationAdapter will look like) 😄

Heh, for future reference that's going to be part of #2380, the relevant commit is https://github.com/mysociety/pombola/commit/eb751abc7f2f8ae115893ff64af965705db7d78a.



> The other thing which I'm just raising to check that you've thought about is that it could potentially get really confusing whether "person" refers to an actual Person or a generic recipient (which might be an Organisation). Is this consistent now, do you think? e.g. is PersonMixin in pombola/writeinpublic/client.py well named or should that be RecipientMixin? (That might not be the best example, but you see what I mean, I hope.)

Hmm yep, good point. I will have a quick look through and change the obvious bits to make them hopefully less confusing.

chrismytton:
I've renamed a couple of things in 74326cb, but it's a bit tricky because I want to keep the naming similar to the WriteInPublic API as much as possible to avoid confusion at that level. Hopefully that change has at least made things a little clearer.

chrismytton:
This is now live.

Issue #2397, opened 2018-03-12
South Africa tests broken on master


chrismytton:
The tests for South Africa are failing on `master`. It looks like switching the attendance API to use https in https://github.com/mysociety/pombola/pull/2390 was the culprit.

PR #2398, opened 2018-03-12
Fix South Africa tests

chrismytton:
Changing the attendance API to use https broke the tests. This changes the tests to use https to fix things again.

Fixes https://github.com/mysociety/pombola/issues/2397


mhl:
:+1: Thanks for fixing this - I was too hasty in merging it :(

PR #2400, opened 2018-03-13
[WriteInPublic] Handle errors when retrieving messages

### LINKED ISSUE #2399: [WriteInPublic] Error on non-contactable committees
When you visit the messages page of a committee that can't be contacted because it doesn't have an email address and therefore isn't in WriteInPublic the following error occurs:

```
Internal Server Error: /organisation/standing-committee-on-public-accounts/messages/ 
Traceback (most recent call last): 
  File "/data/vhost/www.pa.org.za/pombola-virtualenv/lib/python2.7/site-packages/django/core/handlers/base.py", line 132, in get_response 
    response = wrapped_callback(request, *callback_args, **callback_kwargs) 
  File "/data/vhost/www.pa.org.za/pombola-virtualenv/lib/python2.7/site-packages/django/views/generic/base.py", line 71, in view 
    return self.dispatch(request, *args, **kwargs) 
  File "/data/vhost/www.pa.org.za/pombola/pombola/writeinpublic/views.py", line 141, in dispatch 
    return super(WriteInPublicMixin, self).dispatch(*args, **kwargs) 
  File "/data/vhost/www.pa.org.za/pombola-virtualenv/lib/python2.7/site-packages/django/views/generic/base.py", line 89, in dispatch 
    return handler(request, *args, **kwargs) 
  File "/data/vhost/www.pa.org.za/pombola-virtualenv/lib/python2.7/site-packages/django/views/generic/base.py", line 158, in get 
    context = self.get_context_data(**kwargs) 
  File "/data/vhost/www.pa.org.za/pombola/pombola/writeinpublic/views.py", line 269, in get_context_data 
    context['messages'] = self.client.get_messages(uri) 
  File "/data/vhost/www.pa.org.za/pombola/pombola/writeinpublic/client.py", line 93, in get_messages 
    response.raise_for_status() 
  File "/data/vhost/www.pa.org.za/pombola-virtualenv/lib/python2.7/site-packages/requests/models.py", line 834, in raise_for_status 
    raise HTTPError(http_error_msg, response=self) 
HTTPError: 404 Client Error: NOT FOUND 
```

This is because the app makes a request for the messages to this "person" (AKA committee) and WiP returns a 404 because it doesn't recognise the person_id.

I think we should do two things:

1. Handle the `requests.HTTPError` properly
2. Don't link to the messages page for committees that can't be contacted.

chrismytton:
If the committee that we're requesting message for doesn't exist in the WiP instance then it will return a 404. We want to ignore this error and just use an empty list of messages in the context in that case.

Fixes #2399

### DIFF
--- a/pombola/south_africa/views/organisations.py
+++ b/pombola/south_africa/views/organisations.py
@@ -63,6 +63,11 @@ class SAOrganisationDetailView(CommentArchiveMixin, OrganisationDetailView):
         except EmptyPage:
             context['positions'] = paginator.page(paginator.num_pages)
 
+        # Determine if this org is a committee that can be contacted
+        context['contactable_via_writeinpublic'] = \
+            self.object.kind.name == 'National Assembly Committees' \
+            and self.object.contacts.filter(kind__slug='email').exists()
+
         return context
 
     def add_parliament_counts_to_context_data(self, context):
--- a/pombola/writeinpublic/client.py
+++ b/pombola/writeinpublic/client.py
@@ -89,7 +89,12 @@ class WriteInPublic(object):
             'api_key': self.api_key,
             'person__popolo_uri': person_popolo_uri,
         }
-        response = requests.get(url, params=params)
-        response.raise_for_status()
-        messages = response.json()['objects']
-        return [Message(m, adapter=self.adapter) for m in messages]
+        try:
+            response = requests.get(url, params=params)
+            if response.status_code == 404:
+                return []
+            response.raise_for_status()
+            messages = response.json()['objects']
+            return [Message(m, adapter=self.adapter) for m in messages]
+        except requests.exceptions.RequestException as err:
+            raise self.WriteInPublicException(unicode(err))
--- a/pombola/south_africa/templates/core/organisation_detail.html
+++ b/pombola/south_africa/templates/core/organisation_detail.html
@@ -36,7 +36,7 @@
 
   {% endif %}
 
-  {% if object.kind.slug == 'national-assembly-committees' %}
+  {% if contactable_via_writeinpublic %}
     
 
     <div class="contact-actions">
--- a/pombola/south_africa/tests.py
+++ b/pombola/south_africa/tests.py
@@ -1416,6 +1416,29 @@ class SAOrganisationDetailViewTest(WebTest):
 
 
 @attr(country='south_africa')
+class SAOrganisationDetailViewWriteInPublicTest(TestCase):
+    def setUp(self):
+        na_committee_kind = models.OrganisationKind.objects.create(name='National Assembly Committees', slug='national-assembly-committees')
+        self.committee = models.Organisation.objects.create(slug='test-committee', kind=na_committee_kind)
+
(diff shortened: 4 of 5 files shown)


pombola/writeinpublic/client.py
```diff
@@ -89,7 +89,10 @@ def get_messages(self, person_popolo_uri):
...
-        response = requests.get(url, params=params)
-        response.raise_for_status()
-        messages = response.json()['objects']
-        return [Message(m, adapter=self.adapter) for m in messages]
+        try:
+            response = requests.get(url, params=params)
+            response.raise_for_status()
+            messages = response.json()['objects']
+            return [Message(m, adapter=self.adapter) for m in messages]
+        except requests.exceptions.RequestException as err:
```
mhl: It worries me a bit that this is catching any error that might arise from the request, including things we might want to know about like 5xx errors. I think I'd be tempted to do something like:
```
try:
    response = requests.get(url, params=params)
    if response.status == 404:
        return []
    response.raise_for_status()
    messages = response.json()['objects']
    return [Message(m, adapter=self.adapter) for m in messages]
```

Then the calling code in `pombola/writeinpublic/views.py` wouldn't need to be changed as part of this PR...

chrismytton: Ah yes you're quite right, this is being a bit over-eager in the errors it catches! Will change inline with your suggestion :+1:

chrismytton:
@mhl I've pushed a two fixup commits:

1. a34fdbd2b2292ae663306b556ff97700bb45405e changes the `contactable_via_writeinpublic` check to ensure the org is a committee, rather than just checking it has an email address (which any org could have)
2. 692d9fc30991a007658fcda14dd4ae54bed22290 makes the change you suggested to handle 404 errors in the client so the view doesn't need to change.

pombola/south_africa/views/organisations.py
```diff
@@ -63,6 +63,17 @@ def get_context_data(self, **kwargs):
...
+        queryset = models.Organisation.objects.filter(
+            pk=self.object.pk,
+            kind__name='National Assembly Committees',
+            contacts__kind__slug='email'
+        ).distinct()
+        if self.object in queryset:
+            context['contactable_via_writeinpublic'] = True
+        else:
+            context['contactable_via_writeinpublic'] = False
+
```
mhl: Sorry to do that annoying thing of coming back to check fixups and seeing something else in the original PR, but ...

This is a surprising way to write this to me - is there a reason to do it this way rather than, say:
```
        context['contactable_via_writeinpublic'] = \
            self.object.kind.name == 'National Assembly Committees' \
            and self.object.contacts.filter(kind__slug='email').exists()
```

chrismytton: Ahh that's much nicer! I tried something similar but it ended up more verbose because I didn't know about the `.exists()` method. Will make that change now 🙏

chrismytton: @mhl I've fixed that in 688b82e480a6179a6c8065e8633b250b81d408c6

PR #2426, opened 2018-05-30
Openup survey land issues

JenMysoc:
Adding the open up survey to PMG for their new monthly survey plus picture


pombola/south_africa/templates/home.html
```diff
@@ -104,10 +104,10 @@ <h2>
     <div class="wrapper">
         <div class="home__news__row">
             <div class="home__news__survey">
-                <h3>#SAAbailout</h3>
-                <img class="home__news__survey__thumb" src="{% static 'images/SAA survey.jpg' %}" />
-                <h4>SAA: Bail-out or not?</h4>
-                <a href="https://www.pa.org.za/blog/saa-business-rescue-haveyoursay" class="button button--low-case">Have your say</a>
+                <h3>#LandExpropriation</h3>
```
chrismytton: I'm going to remove this `<h3>` because the same text is already in the image, and it looks a bit odd with both! (See below)

screen shot 2018-05-31 at 10 20 01

chrismytton: Done in 34127c8

screen shot 2018-05-31 at 10 48 08

PR #2428, opened 2018-06-11
2427 2017 members interests

xybrnet:
Adds 2017 member's interest data in correct format for import.


mhl:
Looks good :sparkles:  :+1:

Issue #2415, opened 2018-04-16
[ZA] How can we reduce confusion over the “Find your representative” form on the homepage?


zarino:
![screenshot-2018-4-16 people s assembly](https://user-images.githubusercontent.com/739624/38817985-78ebd270-4191-11e8-9d58-e40f5be7d45f.png)

PMG have suggested that people are typing representatives’ names, or general search terms, into the “Find your representative” input on the homepage. Which returns useless location results like this:

![screenshot-2018-4-16 people s assembly 1](https://user-images.githubusercontent.com/739624/38818168-fdbef590-4191-11e8-8cb5-5e30f299fbd9.png)

They’ve suggested changing the title of the form to “Find your constituency office and representatives” (to make “office” more prominent, and give a clue that you’re searching for a location). Or, more radically, removing the search form entirely, and instead linking to it, on a separate page, via a big red button like the “Write to…” buttons further down the page.

They’d like our advice on what might work best.

zarino:
My first reaction was something like this:

![screenshot-2018-4-16 people s assembly 2](https://user-images.githubusercontent.com/739624/38819467-62bff1dc-4193-11e8-82c1-fbaaff7d3278.png)

* Grey text inside inputs should be a “placeholder” _example_ of valid input, not an instruction on what to enter – so, we show an example location, to give people a hint that _they too_ should type in a location.
* Bigger label text

That’s it! Although I do still wonder whether “Find your…” is too ambiguous, and might be misunderstood as meaning “Search for your… by name”. I wonder how else we could word the heading?

zarino:
Just had a thought about a heading for that form. What about "Who represents you?"

JenMysoc:
Ok we have feedback from PMG :)

They would like the following options from @zarino 

 “Who represents you?” to replace “Find Your Representative” 
 “grey text inside inputs should be a “placeholder” example of valid input, not an instruction on what to enter, i.e.  Show an example location, to give people a hint that they too should type in a location; with bigger font

zarino:
Pull request here: https://github.com/mysociety/pombola/pull/2435

PR #2435, opened 2018-06-11
[ZA] Tweak presentation of homepage “Find your rep” form

zarino:
An attempt at fixing #2415, by making it more obvious that you should enter a location into the search box, not an MP’s name.

# Narrow screens

screen shot 2018-06-11 at 17 28 39

# Wide screens

screen shot 2018-06-11 at 17 28 04


mhl:
:+1: Looks good to me. I was wondering if you'd thought about setting up an A/B test for this new version, but I guess it's not obvious how to test in GA that more people are putting in addresses than names after the change!

PR #2443, opened 2018-07-04
Grid layout for Keyna reps

### LINKED ISSUE #2423: [KE] MP/representative list page redesign
Main thing for the rest of the site is to have MPs/Parliamentarians displayed in a picture grid format with their name and area underneath. 

Current look:

screen shot 2018-05-09 at 10 44 13

but from what she's described it sounds more like the Mac file preview grid format is what she wants:

screen shot 2018-05-09 at 12 34 45

zarino:
Fixes #2423.

Sorry, I don‘t have any representative’s photos on my local machine, so there are lots of anonymous placeholder images in the screenshots below. Obviously, on the live site, those images will be replaced by people’s faces :-)

# Narrow screens

screen shot 2018-07-05 at 13 59 33

# Bigger screens

screen shot 2018-07-05 at 14 00 58

The images now lazy-load, meaning the page should be _fairly_ fast to load the first time, and members’ images will only be pulled in as the user scrolls down.

### DIFF
--- a/pombola/core/static/sass/_common.scss
+++ b/pombola/core/static/sass/_common.scss
@@ -29,19 +29,79 @@ ul.ui-autocomplete {
     float: right;
 }
 
-.grid-listing-item {
-    float: left;
-    width: 210px;
-    margin: 10px;
-    padding: 5px;
-    overflow: hidden;
+.grid-listing {
+    overflow: auto;
 
-    span.name {
-        font-size: 120%;
-        display: block;
-        height: 2.5em;
-        overflow: hidden;
+    @include display-flex();
+    @include flex-wrap(wrap);
+
+    @media (min-width: 40em){
+        margin: -10px;
     }
 
-    @include box-shadow(rgba(0, 0, 0, 0.3) 1px 1px 12px);
+    * {
+        box-sizing: border-box;
+    }
+}
+
+.grid-listing-item {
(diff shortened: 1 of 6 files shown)


pombola/core/static/sass/_common.scss
```diff
@@ -29,19 +29,79 @@ ul.ui-autocomplete {
...
+        display: block;
+        margin-top: 0.5em;
     }
 
-    @include box-shadow(rgba(0, 0, 0, 0.3) 1px 1px 12px);
+    .parent-place {
+        display: block;
+        color: #a7a28d;
+        margin-top: 0.1em;
+    }
```
chrismytton: I think the changes in this file need to be made to one that's specific to Kenya (somewhere in `pombola/kenya/static/sass/`) rather in this file, which is shared between countries.

As a bonus that should fix the Travis build, which is currently broken because `pombola/core/static/sass/admin.scss` includes `pombola/core/static/sass/_common.scss` but not `pombola/core/static/sass/_core.scss` so it's complaining about `Undefined mixin 'display-flex'` when trying to build `admin.scss`.

zarino:
@chrismytton Looks like Travis is now happy 👍 Shall I merge and deploy?

chrismytton:
@zarino Woo! Yep, go for it 👍

PR #2444, opened 2018-07-06
Add link to MuniMoney on Rep results page

xybrnet:
@mhl 

This requests extra info from the mapit API, and adds a link to the ward's local municipality on Municipal Money to the Rep Locator results page.

It seems there aren't any tests for this section of code. Can you please confirm that is the case?


xybrnet:
I see the coverage status has decreased, so there are tests for this code.

PR #2467, opened 2018-08-24
Shorten displayed committee names in some contexts

### LINKED ISSUE #2429: [ZA] Change naming of committees on portfolio committee page
PMG have said: On the list of Portfolio Committees:  remove " Portfolio Committee on" so each committee has only its portfolio name, Arts and Culture. 

This can be found both on the WIP Committees page: https://www.pa.org.za/write-committees/recipients/

screen shot 2018-06-11 at 13 59 52

And on the National Assembly committees page: https://www.pa.org.za/committees/

screen shot 2018-06-11 at 14 00 18

jacksonj04:
As per #2429, shortens the displayed committee names in the contexts of writing to committees via Write in Public.

Closes #2429

### DIFF
--- a/pombola/core/migrations/0010_organisation_update_short_name.py
+++ b/pombola/core/migrations/0010_organisation_update_short_name.py
@@ -1,0 +1,38 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+import re
+
+
+def shorten_name(name):
+
+    strip_phrases = [
+        'Portfolio Committee on',
+        'Standing Committee on',
+    ]
+
+    for phrase in strip_phrases:
+        name = re.sub('(?i)' + re.escape(phrase), '', name)
+
+    return name.strip()
+
+
+def resave_organisations(apps, schema_editor):
+    Organisation = apps.get_model('core', 'Organisation')
+    organisations = Organisation.objects.all()
+
+    for organisation in organisations:
+        organisation.short_name = shorten_name(organisation.name)
+        organisation.save()
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('core', '0009_organisation_short_name'),
+    ]
+
+    operations = [
+        migrations.RunPython(resave_organisations),
+    ]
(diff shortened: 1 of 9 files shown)


pombola/core/migrations/0011_organisation_alter_short_name.py
```diff
@@ -0,0 +1,19 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
```
alexsdutton: You could combine this into the previous migration so that the "create field" and "populate field" operations are together.

pombola/core/migrations/0010_organisation_update_short_name.py
```diff
@@ -0,0 +1,38 @@
...
+def resave_organisations(apps, schema_editor):
+    Organisation = apps.get_model('core', 'Organisation')
+    organisations = Organisation.objects.all()
+
+    for organisation in organisations:
+        organisation.short_name = shorten_name(organisation.name)
+        organisation.save()
+
+
+class Migration(migrations.Migration):
```
alexsdutton: You could combine this into the previous migration so that the "create field" and "populate field" operations are together.

pombola/core/models.py
```diff
@@ -767,6 +768,24 @@ def is_ongoing(self):
...
+            'Standing Committee on',
+        ]
+
+        for phrase in strip_phrases:
+            name = re.sub('(?i)' + re.escape(phrase), '', name)
+
+        return name.strip()
+
+    def save(self, *args, **kwargs):
+        self.short_name = self.__shorten_name()
```
alexsdutton: Should this happen only if `not self.short_name`, so that people can tweak it manually?

jacksonj04: I contemplated this approach, but then made `short_name` a non-editable field in favour of maintaining an entirely programmatical approach. The reasoning being that otherwise we would have had to create an entirely new property to access `short_name` based on if it was set or not (and then also created `get_` and `set_` methods for a hidden `_short_name` or similar), which would have made sorting awkward.

Issue #2462, opened 2018-08-17
[KE] Fetch and store SMS messages, then make them available to the homepage template


zarino:
Broken out from #2421.

Chat from Slack earlier:

> They have a service that gathers SMS messages, and exposes a JSON API through which you can extract all the messages. They want a hand-picked selection of these messages to appear on the homepage. I originally thought they’d do all the triage / selection via this SMS service (assuming it has its own admin interface), but _now_ it sounds like it’s more of a dumb pipe, and _we’ll_ have to do that triage / selection ourselves, via the Django admin.

> I’m guessing it’ll mean adding a little app / component / whatever Django calls it, to periodically poll that JSON API for new comments, cache them in the database, and let admins toggle them visible / invisible on the homepage.

> Yeah, that's what i was thinking too. I think this won't be too bad in the django admin. You can add custom actions, so someone could select multiple comments from the list page, and do the "make visible" action.

crowbot:
![](https://github.trello.services/images/mini-trello-icon.png) [MZALENDO: Fetch SMS messages and display them on the homepage](https://trello.com/c/IEsltCWl/1100-mzalendo-fetch-sms-messages-and-display-them-on-the-homepage)

zarino:
Just an extra note, while we’re at it – in addition to the SMS messages themselves, there’s a related piece of text they’ll want to edit, which is the "Question" those SMS messages are meant to be responses to.

![question](https://user-images.githubusercontent.com/739624/45079575-d52afc00-b0ea-11e8-9e8c-f0353a555053.png)

The question will change periodically, and the same person who’s showing/hiding messages will also be updating that question text, so maybe it makes sense for it to be editable in the same part of the Admin UI?

zarino:
After thinking about it, an ajax solution for the _display_ of the messages felt like overkill. I’ve gone with something much simpler in #2478, which uses a URL parameter to paginate messages (as a non-JS fallback) with a sprinkling of JavaScript on top to avoid the page reload for most people.

Right now, in #2478, the messages are all just hard-coded in `kenya/views.py`, but it should be dead simple to get them out of a model, once that model exists.

PR #2481, opened 2018-10-01
[Kenya] Retrieve and store SMS messages

chrismytton:
This adds the ability to fetch and store SMS messages from the API we're using. It also adds an admin section for accepting or rejecting messages, which can then be used to decide what to display on the homepage.

Part of #2462


chrismytton:
This is what the admin looks like currently. You can select multiple messages and then use the dropdown at the top to bulk accept or reject. I've disabled deleting messages (because they'll re-appear when you fetch from the API) and adding new messages (because we only want ones from the API).

image

Issue #2475, opened 2018-09-06
[ZA] Simplify province pages


zarino:
Based on comments from Gaile and Rashaad about "province pages" like https://www.pa.org.za/place/eastern-cape/ …

![screenshot_2018-09-06 people s assembly](https://user-images.githubusercontent.com/739624/45156044-dfc1c000-b1d4-11e8-9509-1e581d5b45cd.png)

* The “People / Constituency Offices” tabs are broken (they don‘t look like tabs at all!) – Megan suggests dropping the “tabs” concept entirely, and just have the people list/accordion always visible, meaning we don’t need a “people” tab at all.
   * Effectively we’d be combining [/place/&lt;name&gt;](https://www.pa.org.za/place/eastern-cape/) and [/place/&lt;name&gt;/places](https://www.pa.org.za/place/eastern-cape/places/) so that people and places are always included on the same page all the time (but folded away nicely in the accordion, as currently).
   * Need to check performance implications of this. Will the server be able to render both sets of data quick enough?
   * Will probably need some developer time, for combining the Django views.
* It’s not made very clear that—for example—“Members of the Provincial Legislature (63)” contains **past** members as well as present members – Megan suggests breaking past/present out into separate accordion sections?
* They want to rename those accordion links so they’re no longer all “Members of…”. New wording:
   * Members of the Provincial Legislature
   * National Assembly members from provincial list
   * Permanent Delegates to National Council of Provinces
* Remove the “Expand all subsections” link and the “There are 121 people related to…” sentence.
* Remove grey lines beside [Constituency Office list items](https://www.pa.org.za/place/eastern-cape/places/).
   * I think they’re there because the template is trying—and failing—to print out the **people** related to that location?

zarino:
#2494 has done some work towards tidying up the bottom half of this page, by fixing the tab styling and removing the summary row:

![screen shot 2018-10-12 at 09 42 10](https://user-images.githubusercontent.com/739624/46858382-1e195300-ce03-11e8-8971-54a9b31fc920.png)

Still to do (as above):

* Show both people and offices on the same page
* Split out historical memberships
* Rename the memberships
* Remove the grey lines beside office list items

chrismytton:
PMG have asked if it's possible to get rid of the whitespace to the right of the Google Map?

zarino:
> PMG have asked if it's possible to get rid of the whitespace to the right of the Google Map?

[...]

PR #2494, opened 2018-10-09
[ZA] Quick fixes for Province page tab styling

zarino:
It turns out combining the [/place/&lt;name&gt;](https://www.pa.org.za/place/eastern-cape/) and [/place/&lt;name&gt;/places](https://www.pa.org.za/place/eastern-cape/places/) views was trickier than anticipated.

So while I wait for @chrismytton to find some time for that, here’s a super quick fix for the broken tab styling and extraneous membership total that the PMG folks wanted fixed as part of #2475.

screen shot 2018-10-09 at 17 46 43

### DIFF
--- a/pombola/core/templatetags/active_class.py
+++ b/pombola/core/templatetags/active_class.py
@@ -6,7 +6,7 @@ register = template.Library()
 
 @register.simple_tag
 def active_class(request_path, name, **kwargs):
-    """ Return the string ' active ' current request.path is same as name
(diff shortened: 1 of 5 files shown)


chrismytton:
Tests were failing because they were doing the following assertion:

```python
self.assertContains(resp, "There are 3 people related to Test Place.")
```

I've removed those assertions now, so going to wait for Travis to run the tests before I review this.

PR #2499, opened 2018-10-29
[KE] [WIP] Opengraph and twitter meta tags

### LINKED ISSUE #2497: [KE] Add twitter and opengraph meta tags to pombola pages and wordpress blog posts
Add the necessary markup for [Twitter Cards](https://developer.twitter.com/en/docs/tweets/optimize-with-cards/overview/abouts-cards) and [Facebook Open Graph ](https://developers.facebook.com/docs/sharing/webmasters/#markup) so that when links to Mzalendo are shared on Twitter it comes up with a preview.

This request came in from Abiud today.

zarino:
Fixes #2497.

WIP


chrismytton:
In the end Mzalendo were happy to just have Opengraph tags on the blog. For future reference I think this would require a rethink of the approach we take with the `<title>` tag before we can get something like this implemented. For now I'm going to close this PR as we're going to be handing these sites off to partners soon and I don't want to leave half-completed work hanging.

PR #2511, opened 2018-11-14
WIP: [ZA] Output relevant data on positions held for easier use downstream

jacksonj04:
This is foundation work to expose data for the [People's Assembly Morph scraper](https://morph.io/everypolitician-scrapers/south-africa-national-assembly) in a clean way without needing to resort to scraping and processing HTML pages.


pombola/core/views.py
```diff
@@ -508,11 +509,113 @@ def position(request, pt_slug, ok_slug=None, o_slug=None):
         context['alphabetical_link_from_query_parameter'] = True
     context['positions'] = positions
 
-    return render_to_response(
-        template,
-        context,
-        context_instance=RequestContext(request)
-    )
+    if request.GET.get('view') == 'csv':
```
chrismytton: My Friday Brain very briefly thought this was referring to a Django View! Perhaps this parameter could be called something like `format` to avoid any ambiguity?

pombola/core/views.py
```diff
@@ -508,11 +509,113 @@ def position(request, pt_slug, ok_slug=None, o_slug=None):
...
+
+        response = HttpResponse(content_type='text/csv')
```
chrismytton: With `text/csv` my browser downloads the file, rather than just displaying it. So while this is technically more _correct_, I wonder if it might be more convenient if the content type was `text/plain`, making it easier to check in the browser?

pombola/core/views.py
jacksonj04: 👍, that's exactly the word my brain wanted to put in there but couldn't find.

pombola/core/views.py
jacksonj04: Given that this is mostly going to feed things behind the scenes, I think the correct format is probably more useful (and may well stop browsers trying to do interesting things with format extensions such as appending a `.txt` on download, which is definitely wrong.

chrismytton: @jacksonj04 Yep, makes sense.

PR #2536, opened 2018-11-30
Do not print empty "related people" list for constituency offices

### LINKED ISSUE #2528: Remove the grey lines beside office list items
This relates to ticket #2475 province redesign.
Estimated time for design - 1 day

zarino:
I _think_ this should fix #2528, but I can’t tell because my local version of people’s assembly won’t render the constituency office page at all (`AttributeError: 'str' object has no attribute 'slug'` from `{% should_display_place organisation_kind as display_place %}` in `/vagrant/pombola/pombola/south_africa/templates/core/person_list_item.html:29`)

@chrismytton care to try it out?

### DIFF
--- a/pombola/south_africa/templates/south_africa/constituency_office_list_item.html
+++ b/pombola/south_africa/templates/south_africa/constituency_office_list_item.html
@@ -19,7 +19,7 @@
       </ul>
     </div>
     <div class="column">
-      {% if not skip_positions %}
+      {% if object.related_positions and not skip_positions %}
         <ul class="unstyled-list">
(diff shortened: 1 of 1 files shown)


chrismytton:
@zarino I'm getting the same error as you locally. Not sure what's causing it though, I will investigate.

chrismytton:
Turns out this wasn't the right fix. The actual problem was the exception itself, which is suppressed in production, hence the grey lines, but manifests itself in development.

Have opened another pull request with the actual fix over in https://github.com/mysociety/pombola/pull/2580.

PR #2543, opened 2019-01-14
[KE] Group speech counts for MPs by year instead of month.

### LINKED ISSUE #2408: [KE] Organise MP speech counts by year and display
From Jessica: 

> [we want] Each MPs speech counts organised per year as we have realised most people don't care about the monthly entries. MPs that have served more than one term have no way of figuring out which years were more productive.

Example of current "Appearances" page for an MP: http://info.mzalendo.com/person/aden-duale/appearances/

We need to do some investigation with Jessica to find out:
How they want this to be displayed (eg as a number on the Reps profiles or as something else)
Where on the profile they'd want it to go (in the sidebar, in the box about parliamentary appearances). Finally we'd need to know if they want text by it and if that text might change. 

How do we currently calculate the speech counts - is this the script we run every year? If so we also need to know if it's still done once a year, but instead showing a yearly rather than monthly total?

jacksonj04:
Closes #2408.

### DIFF
--- a/pombola/hansard/models/entry.py
+++ b/pombola/hansard/models/entry.py
@@ -24,6 +24,20 @@ class EntryQuerySet(models.query.QuerySet):
 
         return counts
 
+    def yearly_appearance_counts(self):
+        """Return an list of dictionaries for dates and counts for each year"""
+
+        # would prefer to do this as a single query but I can't seem to make the ORM do that.
+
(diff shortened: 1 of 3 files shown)


jacksonj04:
"Month" now corrected to "year" in the copy!

chrismytton:
Awesome, thanks @jacksonj04!

PR #2547, opened 2019-02-01
[Kenya] Add speaker filter to hansard search

### LINKED ISSUE #2420: [KE] Improve ability to search hansard
screen shot 2018-05-09 at 10 00 59

Jessica wants to make it easier to search hansard for specific topics and see which MPs are speaking about them. This is linked to #2408 in a way in that if we're pulling out the topics from hansard for word clouds maybe we could add either a search by topic or suggest topics to search by. It would be ideal if the topics returned would also break it down by person (so you could click on an MP and see what they had said).

chrismytton:
This adds an extra field to the Hansard search form which allows you to put in a speaker's name to filter by. To do this I've added an extra field into the Hansard entry search index which contains the name of the speaker from Hansard, plus their name in our database and any known aliases.

Fixes #2420 

## Note to merger/deployer

When this is deployed we will need to run `./manage.py rebuild_index` which will rebuild the elasticsearch index to include the new speaker name field.

# Screenshots

## No speaker filter

screenshot 2019-02-01 at 16 12 53

## With speaker filter

screenshot 2019-02-01 at 16 13 08

### DIFF
--- a/pombola/search/templates/search/hansard.html
+++ b/pombola/search/templates/search/hansard.html
@@ -28,6 +28,11 @@
                 <option value="adate"{% if order == 'adate' %} selected{% endif %}>Oldest first</option>
               </select>
 
+              <div>
+                <label for="speaker-name">Speaker name</label>
+                <input id="speaker-name" type="text" name="speaker" type="text" value="{{ speaker }}">
+              </div>
+
             <input type="submit" value="Search" class="button">
         </div>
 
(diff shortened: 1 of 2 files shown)


chrismytton:
> Re the deployer comment - index rebuild does happen automagically as well, doesn't it?

When _new_ items are added to the database they automatically get added to the index, yeah. But because this is adding a new field to an existing index we need to rebuild it to reindex all the existing Hansard entries.

PR #2581, opened 2019-03-19
[ZA] Set number of members in the assembly and NCOP to be static

### LINKED ISSUE #2556: [ZA] Change MP number on parliaments page to be static
From Megan:

> Can the MP numbers on the parliament page https://www.pa.org.za/organisation/is/parliament/ be changed to static numbers on the pictures - 400 for the National Assembly and 90 for NCOP - those current numbers are misleading.

jacksonj04:
Adds a new `seats` field to `Organisation`s, representing the nominal number of seats. Displays this instead of the number of position holders (which has renamed variables elsewhere for clarity).

Closes #2556 

---

image

### DIFF
--- a/pombola/surveys/migrations/0002_auto_20190319_1101.py
+++ b/pombola/surveys/migrations/0002_auto_20190319_1101.py
@@ -1,0 +1,18 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('surveys', '0001_initial'),
+    ]
+
+    operations = [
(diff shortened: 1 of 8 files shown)


pombola/surveys/migrations/0002_auto_20190319_1101.py
```diff
@@ -0,0 +1,18 @@
...
+    dependencies = [
+        ('surveys', '0001_initial'),
+    ]
+
+    operations = [
+        migrations.AlterModelOptions(
+            name='survey',
+            options={'get_latest_by': 'created'},
+        ),
+    ]
```
chrismytton: Oops, think this missing migration was my fault, sorry!

PR #2613, opened 2019-06-06
Split into Kenya

jacksonj04:
**DO NOT MERGE**

This PR is currently so that analysis and tests can run. Merging it will cause **Very Bad Things** to happen.

Part of https://github.com/mysociety/pombola/issues/2630


jacksonj04:
This branch has been pushed to `master` in https://github.com/mysociety/pombola-ke. All changes relating to Kenya should now be made there.

PR #2643, opened 2019-07-02
Add position column to minister attendance table

xybrnet:
This adds a position column to the table when viewing attendance for ministers in order to differentiate between ministers and deputies.

When we retrieve the ministerial positions, we sort by person, then by start date, in order to group by person, and have their latest position last when iterating over the list and setting their position in the attendance records.


chrismytton:
@xybrnet This is now live, looking good! https://www.pa.org.za/mp-attendance/

xybrnet:
Great, thank you @chrismytton!

PR #2647, opened 2019-07-03
2646 pre and post 2019 election attendance

xybrnet:
This is a bit ugly, but we need to differentiate 2019 attendance between pre- and post-election records, the dividing date being the opening of the 6th parliament, 1 July 2019.

The PMG API returns annual data from 1 Jan - 31 Dec, so if the user has selected to view records for 2019 (which is the default view for the page until there we receive records for 2020), we divide the results into those before and after 1 July 2019 return the selected records.

### DIFF
--- a/pombola/south_africa/tests.py
+++ b/pombola/south_africa/tests.py
@@ -836,7 +836,7 @@ class SAMpAttendancePageTest(TestCase):
             u'meetings_by_member': [
                 {u'member': {
                     u'party_id': 1, u'pa_url': u'http://www.pa.org.za/person/person1/',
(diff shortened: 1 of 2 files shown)


xybrnet:
Thank you for catching that @chrismytton.
I've updted first commit to include the fix.

chrismytton:
> Thank you for catching that @chrismytton.
> I've updted first commit to include the fix.

Thanks! Will take another look at this today.

chrismytton:
@xybrnet This has now been deployed to the live site: https://www.pa.org.za/mp-attendance/

xybrnet:
Thank you for pushing this out @chrismytton!

PR #2657, opened 2019-07-16
2656 import 2018 members interests

xybrnet:
This is the 2018 member's interests data that is in a format which is ready to import.

There are some instructions for doing so here:
https://github.com/mysociety/pombola/tree/master/pombola/south_africa/data/members-interests

I had to change some of the fontspec values for the pdf scraper to function correctly, and ended up cleaning it up a bit to be able to see what happens in there.

I tested the changes by running it on 2017 data, which gave the same output as the prior version.


chrismytton:
@xybrnet This has been deployed and the above command has been run, so the 2018 members interests data should now be loaded 🎉

xybrnet:
🎉

PR #2659, opened 2019-07-18
Add an output for organisation memberships in Popolo format

jacksonj04:
Based on the [committees JSON output](https://www.pa.org.za/api/committees/popolo.json) which currently feeds WriteInPublic.

## To do:

* [ ] Double check which ID WiP expects to see.

### DIFF
--- a/pombola/core/views.py
+++ b/pombola/core/views.py
@@ -22,7 +22,7 @@ import django
 from django.core.urlresolvers import reverse
 from django.db.models import Count
 from django.db.models import Q
-from django.http import HttpResponse
+from django.http import HttpResponse, JsonResponse
 from django.shortcuts  import render_to_response, get_object_or_404, redirect
 from django.template   import RequestContext
 from django.views.decorators.cache import cache_control, never_cache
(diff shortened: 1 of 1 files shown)


pombola/core/views.py
```diff
@@ -611,6 +611,31 @@ def handle_approx_date(date):
 
         return response
 
+    elif request.GET.get('format') == 'popolo':
+
+        persons = []
+
+        for position in positions:
```
chrismytton: This API seems quite slow at the moment. Perhaps we could filter down the list of positions and fetch the related people to speed things up?

```python
positions.filter(person__contacts__kind__slug='email').select_related('person')
```

pombola/core/views.py
```diff
@@ -611,6 +611,31 @@ def handle_approx_date(date):
...
+                    {
+                        'id': request.build_absolute_uri(person.get_absolute_url())
+                                    .replace('http://', 'https://'),
+                        'name': person.name,
+                        'email': email_query.first().value,
+                        'contact_details': []
+                    }
+                )
+
+        response = JsonResponse({'persons': persons})
```
chrismytton: Looks like this is currently returning duplicates where people hold more than one position. I think that's because a person can hold multiple positions, so the same person appears more than once in the list of positions.

chrismytton:
Closing in favour of https://github.com/mysociety/pombola/pull/2661 and subsequent tweaks.

PR #2665, opened 2019-07-31
Improve MP assets filter interaction

xybrnet:
A few improvements to improve user interaction with the filters on the MP assets list page.
https://www.pa.org.za/interests/

Different ways to group the data was separated from the filters to make that fact more obvious.
e.g. Number of declarations per MP. A user now selects the way tp group the results first, and then filters those results.

Only the appropriate categories are returned when viewing the numbers by source.

image


chrismytton:
This is now live :tada: https://www.pa.org.za/interests/

PR #2682, opened 2019-09-19
Only display ministers without a party, when not filtered by party

xybrnet:
@chrismytton 

Some previous ministers are no longer members of parties. We need to check for this before trying to access an element in the parties list.

We also only display these ministers when the page is not filtered by party.


chrismytton:
This has now been deployed to the live site: https://www.pa.org.za/mp-attendance/

